PR #96, opened 2017-01-20
only use production User agent str in production

ccemeraldeyes:


ccemeraldeyes:
@johnsonj should have asked for the review before the merge but I tested and got over-excited that it worked.

brokerapi/brokers/models/service_broker.go
```diff
@@ -153,7 +153,11 @@ var (
 // This custom user agent string is added to provision calls so that Google can track the aggregated use of this tool
 // We can better advocate for devoting resources to supporting cloud foundry and this service broker if we can show
 // good usage statistics for it, so if you feel the need to fork this repo, please leave this string in place!
-var CustomUserAgent = "cf-gcp-service-broker 1.0.0"
+var CustomUserAgent = "cf-gcp-service-broker-test 1.0.0"
```
johnsonj: i like this pattern!

i recommend making this a local var to the package and making a `UserAgent()` fn to return it, but I am paranod.

PR #109, opened 2017-02-20
Concourse pipeline

ccemeraldeyes:


ccemeraldeyes:
@evandbrown please review. I can't seem to assign to you.

ci/pipeline.yml
```diff
@@ -0,0 +1,119 @@
...
+    type: docker-image
+    source:
+      repository: frodenas/gcs-resource
+
+resources:
+  - name: gcp-service-broker
+    type: git
+    source:
+      uri: https://github.com/GoogleCloudPlatform/gcp-service-broker.git
+      branch: concourse
```
evandbrown: Is the `concourse` branch like a `develop` branch? We use `develop` and `master` in other projects.

ci/pipeline.yml
```diff
@@ -0,0 +1,119 @@
...
+
+  - name: version
+    type: semver
+    source:
+      bucket: gcp-service-broker-semver
+      key: semver
+      endpoint: storage.googleapis.com
+      access_key_id: {{storage_key_id}}
+      secret_access_key: {{storage_secret_key}}
+      initial_version: 3.1.3
```
evandbrown: Is this really the version we're starting from?

ci/pipeline.yml
```diff
@@ -0,0 +1,119 @@
...
+resources:
+  - name: gcp-service-broker
+    type: git
+    source:
+      uri: https://github.com/GoogleCloudPlatform/gcp-service-broker.git
+      branch: concourse
+      ignore_paths:
+        - ci/*
+
+  - name: gcp-service-broker-ci
```
evandbrown: Why 2 resources for the same git repo/branch? I see one ignores the `ci/` dir and the other doesn't. Possible to use just 1 and avoid cloning twice?

ci/pipeline.yml
```diff
@@ -0,0 +1,119 @@
...
+      endpoint: storage.googleapis.com
+      access_key_id: {{storage_key_id}}
+      secret_access_key: {{storage_secret_key}}
+      initial_version: 3.1.3
+
+jobs:
+  - name: test-unit
+    plan:
+      - aggregate:
+        - get: gcp-service-broker
```
evandbrown: This is aggregating the two resources that are the same git repo. Is that necessary?

ci/pipeline.yml
ccemeraldeyes: yes, once this is merged I'll change the branch to master.

ci/pipeline.yml
ccemeraldeyes: it's essentially a way to separate out the ci config as a resource. The fact that it's in the same repo is just how we've been setting things up, but really the code and the ci config should be considered different resources (which also allows for out-of-sync updating).

ci/pipeline.yml
ccemeraldeyes: yup! One later than the version that's public now.

ci/pipeline.yml
ccemeraldeyes: yes, so the CI one is independent.

PR #112, opened 2017-02-24
very basic golang bindings example TAKE 2

ccemeraldeyes:


erjohnso:
Just a few general comments:

- Document the other use-cases in the README, or somehow mention that there is more to it than just GCS
- Indicate that users will need to have the various API's enabled in GCP
- Question: if an API is *not* enabled, will the error you catch and message be meaningful?
- TODO(ccemeraldeyes): Link to the sample from the main README

Otherwise, LGTM!  Thanks for putting this together.

ccemeraldeyes:
* Done
* The api enablement step is part of setting up the service broker, I don't think it's worth mentioning here
* But yes, those errors are surfaced exactly as they are, which look like "you need to enable this api"
* The examples folder is linked from the main readme

PR #127, opened 2017-03-30
Delete async services

ccemeraldeyes:


integration_tests/async_integration_test.go
```diff
@@ -251,7 +258,14 @@ var _ = Describe("AsyncIntegrationTests", func() {
...
 			Expect(err).NotTo(HaveOccurred())
-			pollForMaxFiveMins(gcpBroker, "integration_test_instance")
+			err = pollForMaxFiveMins(gcpBroker, "integration_test_instance")
+			Expect(err).NotTo(HaveOccurred())
+
+			instance := models.ServiceInstanceDetails{}
+			if err := db_service.DbConnection.Unscoped().Where("ID = ?", "integration_test_instance").First(&instance).Error; err != nil {
+				panic("error checking for service instance details: " + err.Error())
+			}
+			Expect(instance.DeletedAt).NotTo(Equal(nil))
```
danawillow: Any reason why this is `Equal(nil)` and the sql one is `BeNil()`?

brokerapi/brokers/cloudsql/broker.go
```diff
@@ -421,20 +421,15 @@ func (b *CloudSQLBroker) Unbind(creds models.ServiceBindingCredentials) error {
...
 func (b *CloudSQLBroker) PollInstance(instanceId string) (bool, error) {
-	var op models.CloudOperation
-	var ii InstanceInformation
 	var instance models.ServiceInstanceDetails
 
 	if err := db_service.DbConnection.Where("id = ?", instanceId).First(&instance).Error; err != nil {
 		return false, models.ErrInstanceDoesNotExist
 	}
 
-	if err := json.Unmarshal([]byte(instance.OtherDetails), &ii); err != nil {
```
danawillow: Why did these get removed?

integration_tests/async_integration_test.go
ccemeraldeyes: Because it shouldn't be Equal(nil).... will fix.

brokerapi/brokers/spanner/broker.go
```diff
@@ -246,12 +282,32 @@ func (s *SpannerBroker) Deprovision(instanceID string, details models.Deprovisio
 	})
 
 	if err != nil {
-		return fmt.Errorf("Error creating client: %s", err)
+		deleteErr := createFakeDeleteOperation(instanceID, instance.ServiceId, err.Error())
+		if deleteErr != nil {
+			return fmt.Errorf("Error deleteing instance: %s. Then encountered error saving operation information: %s", err, deleteErr)
```
danawillow: typo: s/deleteing/deleting/

brokerapi/brokers/cloudsql/broker.go
ccemeraldeyes: Because it *should* be the same to look up by operation create time vs. operation id, and using create time is more consistent

brokerapi/brokers/spanner/broker.go
```diff
@@ -246,12 +282,32 @@ func (s *SpannerBroker) Deprovision(instanceID string, details models.Deprovisio
 	})
 
 	if err != nil {
-		return fmt.Errorf("Error creating client: %s", err)
+		deleteErr := createFakeDeleteOperation(instanceID, instance.ServiceId, err.Error())
```
danawillow: Why is it a fake operation?

ccemeraldeyes: Because Spanner doesn't really use operations for delete - it runs deletes synchronously. But CF doesn't know services can be half async, so I have to pretend it's entirely async. And it simplifies the flow to use a fake operation entry to do that. Will add a comment to clarify.

brokerapi/brokers/gcp_service_broker.go
```diff
@@ -396,18 +396,30 @@ func (gcpBroker *GCPServiceBroker) LastOperation(instanceID string) (models.Last
 		return models.LastOperation{}, models.ErrInstanceDoesNotExist
 	}
 
-	if gcpBroker.ServiceBrokerMap[instance.ServiceId].Async() {
+	if gcpBroker.ServiceBrokerMap[instance.ServiceId].ProvisionsAsync() || gcpBroker.ServiceBrokerMap[instance.ServiceId].DeprovisionsAsync() {
```
johnsonj: these branches are getting pretty deep. Recommend keeping the branch but pullung them out to their own local/private functions

eg:

```go
if async {
   return gcpBroker.lastOperationAsync()
} else {
   return gcpBroker.lastOperationSync()
}
```

brokerapi/brokers/gcp_service_broker.go
```diff
@@ -396,18 +396,30 @@ func (gcpBroker *GCPServiceBroker) LastOperation(instanceID string) (models.Last
...
 		} else {
 
 			if done {
+				// no error and we're done! Delete from the SB database if this was a delete flow and return success
+				deleteFlow, err := gcpBroker.ServiceBrokerMap[instance.ServiceId].LastOperationWasDelete(instanceID)
+				if err != nil {
+					return models.LastOperation{State: models.Succeeded, Description: ""}, fmt.Errorf("Couldn't determine if provision or deprovision flow, this may leave orphaned resources, contact your operator for cleanup")
+				}
+				if deleteFlow {
+					err = db_service.SoftDeleteInstanceDetails(instanceID)
```
johnsonj: did you consider moving this into `PoolInstance`? That way you don't need to expose `LastOperationWasDelete`

brokerapi/brokers/spanner/broker.go
```diff
@@ -235,7 +245,7 @@ func (s *SpannerBroker) Deprovision(instanceID string, details models.Deprovisio
 	}
 
 	// set up client
-	client, err := googlespanner.NewInstanceAdminClient(context.Background(), option.WithUserAgent(models.CustomUserAgent))
+	client, err := googlespanner.NewInstanceAdminClient(context.Background())
```
johnsonj: discussed offline

brokerapi/brokers/spanner/broker.go
```diff
@@ -246,17 +256,24 @@ func (s *SpannerBroker) Deprovision(instanceID string, details models.Deprovisio
...
-		return fmt.Errorf("Error creating client: %s", err)
+		return fmt.Errorf("Error deleting instance: %s", err)
 	}
 
 	return nil
 }
 
-// Indicates that Spanner uses asynchronous provisioning
-func (s *SpannerBroker) Async() bool {
+/// Indicates that Spanner uses asynchronous provisioning
```
johnsonj: nit: extra /

integration_tests/async_integration_test.go
```diff
@@ -207,6 +214,8 @@ var _ = Describe("AsyncIntegrationTests", func() {
 			err = pollForMaxFiveMins(gcpBroker, "integration_test_instance")
 			Expect(err).NotTo(HaveOccurred())
 
+			println("provision")
```
johnsonj: revert debug statement

brokerapi/brokers/gcp_service_broker.go
ccemeraldeyes: talked offline about why not.

brokerapi/brokers/gcp_service_broker.go
```diff
@@ -433,6 +406,37 @@ func (gcpBroker *GCPServiceBroker) LastOperation(instanceID string) (models.Last
 
 }
 
+func (gcpBroker *GCPServiceBroker) lastOperationAsync(instanceId, serviceId string) (models.LastOperation, error) {
+	done, err := gcpBroker.ServiceBrokerMap[serviceId].PollInstance(instanceId)
+	if err != nil {
```
johnsonj: looking good. If you want to go further, I recommend trying out initializing a models.LastOperation at the top and an empty err (perhaps named different?). Set the value of LastOperation.State in the branch as soon as you have determined the state.

Issue #113, opened 2017-02-24
Update Cloud Storage plans


ccemeraldeyes:
Add Coldline and separate out Regional and Multi-Regional

PR #136, opened 2017-04-18
added new storage plans to manifest

rambleraptor:
#113 

Couple questions on this:

1) What tests should I be running on this?
2) Should I be altering the fakes file or the tile.yml file?


ccemeraldeyes:
1. You should try provisioning one of each service plan (this is fine to do once manually) and verifying the right thing gets created in the console.
2. yes to tile.yml. Fakes doesn't need to worry about it, that's just for testing.

ccemeraldeyes:
Also standard and DRA should be removed

Issue #139, opened 2017-07-05
Command to dump current plan information


ccemeraldeyes:
Child issue to #138 

Add `get_plan_info` to the `cmd` folder for users who are running the service broker as an app and may wish to retain their plan GUIDs in the face of the future requirement for them being written to the manifest.

Update README.

PR #144, opened 2017-07-10
139 - add command to dump current plan information

ccemeraldeyes:
#139 - plan information so that ids can be copied to manifest.yml

### DIFF
--- a/db_service/setup_db.go
+++ b/db_service/setup_db.go
@@ -35,6 +35,11 @@ func SetupDb(logger lager.Logger) *gorm.DB {
 	dbHost := os.Getenv("DB_HOST")
 	dbUsername := os.Getenv("DB_USERNAME")
 	dbPassword := os.Getenv("DB_PASSWORD")
+
+	if dbPassword == "" || dbHost == "" || dbUsername == "" {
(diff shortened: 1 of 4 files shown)


db_service/setup_db.go
```diff
@@ -35,6 +35,11 @@ func SetupDb(logger lager.Logger) *gorm.DB {
 	dbHost := os.Getenv("DB_HOST")
 	dbUsername := os.Getenv("DB_USERNAME")
 	dbPassword := os.Getenv("DB_PASSWORD")
+
+	if dbPassword == "" || dbHost == "" || dbUsername == "" {
```
ccemeraldeyes: legit question but not actually relevant to this particular CL.

docs/customization.md
```diff
@@ -0,0 +1,62 @@
+# Installation Customization
```
ccemeraldeyes: yeah, I forgot to update the refactor branch. Should be the proper diff now.

db_service/setup_db.go
johnsonj: this change is probably suited for it's own cleanup CL, it's a positive change but it's not related

PR #148, opened 2017-07-10
fixes #147 - require plans to have id set

### LINKED ISSUE #147: Require plan GUIDs

ccemeraldeyes:
#147

### DIFF
--- a/brokerapi/brokers/brokers_test.go
+++ b/brokerapi/brokers/brokers_test.go
@@ -193,6 +193,13 @@ var _ = Describe("Brokers", func() {
 				}
 			}
 		})
+
+		It("should error if plan ids are not supplied", func() {
+			os.Setenv("PRECONFIGURED_PLANS", fakes.PlanNoId)
+			_, err := brokers.New(logger)
+			Expect(err).To(HaveOccurred())
+			os.Setenv("PRECONFIGURED_PLANS", fakes.PreconfiguredPlans)
+		})
 	})
 
 	Describe("updating broker catalog", func() {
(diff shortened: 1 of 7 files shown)


brokerapi/brokers/brokers_test.go
```diff
@@ -193,6 +193,13 @@ var _ = Describe("Brokers", func() {
 				}
 			}
 		})
+
+		It("should error if plan ids are not supplied", func() {
+			os.Setenv("PRECONFIGURED_PLANS", fakes.PlanNoId)
+			_, err := brokers.New(logger)
+			Expect(err).To(HaveOccurred())
+			os.Setenv("PRECONFIGURED_PLANS", fakes.PreconfiguredPlans)
```
johnsonj: this shouldn't be in the test. I'm guessing you're doing it here because subsequent tests fail? if that's the case this should be getting set in a BeforeEach(..)

brokerapi/brokers/gcp_service_broker.go
```diff
@@ -494,10 +494,9 @@ func getStaticPlans() (map[string][]models.ServicePlan, error) {
 	for _, p := range plans {
 		serviceId := p["service_id"].(string)
 		planName := p["name"].(string)
-
-		id, err := db_service.GetOrCreatePlanId(planName, serviceId)
-		if err != nil {
-			return map[string][]models.ServicePlan{}, err
+		planId, planIdOk := p["id"]
+		if !planIdOk {
+			return map[string][]models.ServicePlan{}, errors.New("Error: plan id required")
```
johnsonj: nit: Error: isn't helpful here, same as in getDynamicPlans

brokerapi/brokers/brokers_test.go
ccemeraldeyes: it's in a before each, I put that there while I was debugging and forgot to take it out. Thanks! Good catch!

PR #151, opened 2017-07-12
149

ccemeraldeyes:
#149


brokerapi/brokers/gcp_service_broker.go
```diff
@@ -202,6 +202,21 @@ func (gcpBroker *GCPServiceBroker) Services() []models.Service {
 	return *gcpBroker.Catalog
 }
 
+func (gcpBroker *GCPServiceBroker) GetPlanFromId(serviceId, planId string) (models.ServicePlan, error) {
+
+	cat := *gcpBroker.Catalog
```
johnsonj: real nit but is this necessary? can't do range(*gcpBroker.Catalog)?

PR #153, opened 2017-07-13
152 - remove storage of credentials on server by passing jwt.Config

ccemeraldeyes:
#152


brokerapi/brokers/bigquery/broker.go
```diff
@@ -59,7 +60,7 @@ func (b *BigQueryBroker) Provision(instanceId string, details models.ProvisionDe
 		params["name"] = name_generator.Basic.InstanceName()
 	}
 
-	service, err := googlebigquery.New(b.Client)
+	service, err := googlebigquery.New(b.HttpConfig.Client(context.Background()))
```
johnsonj: did these services not have the `models.CustomUserAgent` before? I had assumed that got set on the shared client object, but I also had assumed that was used by all the APIs.

ccemeraldeyes: the user agent is either set on the service (as here, see line 67) or in ClientOptions, depending on which format the constructor uses.

ccemeraldeyes:
> Can you centralize the custom user agent assignment?

If you mean integrate it into the http.Client, not easily. I looked into this when this functionality needed fixing for some apis. It involves implementing a RoundTripper and customizing headers and it's just not worth the effort at this point.

danawillow:
Looks like support for application default credentials was removed. Is it possible for that to break anybody? Or does the new stuff handle it automatically?

ccemeraldeyes:
@danawillow 

> Looks like support for application default credentials was removed. Is it possible for that to break anybody? Or does the new stuff handle it automatically?

Yeah, that's the point of this PR, is to remove that and replace it with `Client` or `TokenSource`

Issue #158, opened 2017-07-25
add envconfig repo


ccemeraldeyes:
"github.com/kelseyhightower/envconfig"

better for processing env variables.

PR #159, opened 2017-07-25
158

ccemeraldeyes:
#158 - also updates other libraries and corresponding code changes


evandbrown:
Could you ammend 36ffbc8 and change the commit message to something like "update vendor packages"? Makes it easy to spot vendor changes (and for a minute I thought the change was updating a library called `wheee`)

evandbrown:
Otherwise lgtm

Issue #9, opened 2016-09-27
Default to minimum necessary permissions on bind


ccemeraldeyes:
Use curated roles as much as possible, defaulting to a viewer-type role.

mgoddard-pivotal:
I'm not sure if this is the spirit of this issue, but having a reasonable default for the values passed in on binding of the service seems like a great idea.  For example, if you commonly bind PubSub with the following, how about making that the default:

  `cf bs app_name service_instance_name -c '{"role": "pubsub.admin" }'`

For Spring Cloud Dataflow applications, where this binding happens in the code, it would be handy to not need to know the service-specific JSON snippet to pass in.

PR #161, opened 2017-07-26
If role is not provided, use default role.

felixfriedrich:
Because of https://github.com/cloudfoundry/cli/issues/1173 this pull request sets 'owner' as the default role if no role is specified in the bind-service command.

I guess reading the default role from the config would be great, but I would need some advice how to implement that.

If you don't want to merge it as is, please provide advice on how to improve the code.

Thanks!

### DIFF
--- a/brokerapi/brokers/account_managers/service_account_manager.go
+++ b/brokerapi/brokers/account_managers/service_account_manager.go
@@ -42,7 +42,7 @@ type ServiceAccountManager struct {
 func (sam *ServiceAccountManager) CreateAccountInGoogle(instanceID string, bindingID string, details models.BindDetails, instance models.ServiceInstanceDetails) (models.ServiceBindingCredentials, error) {
 	role, ok := details.Parameters["role"].(string)
 	if !ok {
-		role = "owner"
+		role = "storage.objectAdmin"
 	}
 
 	someName := ServiceAccountName(bindingID)
--- a/brokerapi/brokers/account_managers/service_account_manager.go
+++ b/brokerapi/brokers/account_managers/service_account_manager.go
@@ -19,7 +19,6 @@ package account_managers
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"gcp-service-broker/brokerapi/brokers/models"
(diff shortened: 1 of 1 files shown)


felixfriedrich:
I changed the role from 'owner' to 'storage.objectAdmin'. (I assued owner, means owner of the bucket.) Yet, the role is applied the complete project, not just to the bucket, which would be better.

ccemeraldeyes:
I'd like to see roles default to minimum permissions (with user-provided overrides, a feature that is coming as soon as a large-scale refactor is done), and I'd like it to see it set up for every service. The issue for that is https://github.com/GoogleCloudPlatform/gcp-service-broker/issues/9

Is this a blocker for you? How urgent?

felixfriedrich:
In a case where the minimum permissions would not be good enough and the user provided override is provided via `-c {"role": "overrrideRole"}` the solution you outlined would not solve my problem as arguments to the bind command are not possible in the manifest (see: https://github.com/cloudfoundry/cli/issues/1173).

I'd suggest putting the default in the config of service broker, in the service definition?

Does that make sense? Or did I misunderstand you?

felixfriedrich:
I signed the CLA by the way.

ccemeraldeyes:
> I'd suggest putting the default in the config of service broker, in the service definition?

Yes, that's coming. I'd *estimate* 1 month to delivery. Can you wait that long? Or is this a more urgent blocker than that?

felixfriedrich:
That sounds great, @ccemeraldeyes ! I can wait since the small change in this pull request unblocks me for now. Thanks for helping, I am going to close the pull request.

Best regards from Berlin, Germany,

Felix

felixfriedrich:
One last question though: The roles being specified will they be only applied to the bucket or to the project (as it is at the moment AFAIK)? Should this maybe be configurable as well?

ccemeraldeyes:
It should only apply to the bucket. There's a ticket for that as well that's high priority.

felixfriedrich:
Thanks, I am quite excited about all the changes :-) Thank you!

PR #162, opened 2017-07-26
157 - refactor service and plan env vars

ccemeraldeyes:
Also added service/plan validation via a new lib.

Separating this info out per service will allow easier enable/disable/config of services especially for tile users, and reduces code complexity.

### DIFF
--- a/brokerapi/brokers/gcp_service_broker.go
+++ b/brokerapi/brokers/gcp_service_broker.go
@@ -42,6 +42,7 @@ import (
 	"gcp-service-broker/db_service"
 	"gcp-service-broker/utils"
 	"golang.org/x/oauth2/jwt"
+	"gopkg.in/validator.v2"
 )
 
 type GCPServiceBroker struct {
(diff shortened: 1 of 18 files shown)


brokerapi/brokers/gcp_service_broker.go
```diff
@@ -485,95 +486,25 @@ func GetCredentialsFromEnv() (models.GCPCredentials, error) {
...
-		s.Plans = servicePlans[s.ID]
-
-		if len(s.Plans) > 0 {
-			serviceList = append(serviceList, s)
+		var svc models.Service
+		if err := json.Unmarshal([]byte(os.Getenv(varname)), &svc); err != nil {
+			return []models.Service{}, err
+		} else {
+			if errs := validator.Validate(svc); errs != nil {
+				return []models.Service{}, errs
```
mkjelland: Is it possible to return nil instead of an empty list? The empty list makes me think that this is going to initialize successfully, but with no services.

ccemeraldeyes: it's consistent with the rest of the code base to return an empty list. The api will only look at the error, though.

ccemeraldeyes:
> How do these env variables get set when you deploy as an app? Do they need to be set in manifest.yml with the GOOGLE_SERVICE_NAME name too?

yes.

utils/utils.go
```diff
@@ -28,9 +28,15 @@ import (
...
-	if err := json.Unmarshal([]byte(serviceStr), &serviceList); err != nil {
-		return idToNameMap, fmt.Errorf("Error unmarshalling service list %s", err)
+
+	for _, varname := range models.ServiceEnvVarNames {
+
+		var svc models.Service
+		if err := json.Unmarshal([]byte(os.Getenv(varname)), &svc); err != nil {
+			return map[string]string{}, err
+		} else {
+			serviceList = append(serviceList, svc)
```
johnsonj: can you build the idToNameMap here instead of re-iterating it on the next block?

johnsonj:
should this change include an update to manifest.yml?

ccemeraldeyes:
@johnsonj 
> should this change include an update to manifest.yml?

It does. Are you implying it shouldn't? Or did you miss the change (it's large, so collapsed by default)

johnsonj:
I missed it! thanks

PR #163, opened 2017-07-27
154

ccemeraldeyes:
#154 

refactor/simplify broker config.

### DIFF
--- a/brokerapi/brokers/config/broker_config.go
+++ b/brokerapi/brokers/config/broker_config.go
@@ -1,0 +1,28 @@
+// Copyright the Service Broker Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////////
+//
+
+package config
+
+import (
+	"gcp-service-broker/brokerapi/brokers/models"
+	"golang.org/x/oauth2/jwt"
+)
+
+type BrokerConfig struct {
+	Catalog    map[string]models.Service
+	HttpConfig *jwt.Config
+}
--- a/brokerapi/brokers/config/broker_config.go
+++ b/brokerapi/brokers/config/broker_config.go
@@ -33,7 +33,7 @@ type BrokerConfig struct {
 	ProjectId  string
 }
 
-func NewBrokerConfig() (*BrokerConfig, error) {
+func NewBrokerConfigFromEnv() (*BrokerConfig, error) {
 	var err error
 	bc := BrokerConfig{}
 	creds, err := bc.GetCredentialsFromEnv()
--- a/brokerapi/brokers/config/broker_config.go
+++ b/brokerapi/brokers/config/broker_config.go
@@ -92,14 +92,3 @@ func (bc *BrokerConfig) InitCatalogFromEnv() (map[string]models.Service, error) 
 
 	return serviceMap, nil
 }
-
-func (bc *BrokerConfig) ServiceList() *[]models.Service {
-	svcs := []models.Service{}
-
-	for _, svc := range bc.Catalog {
-		svcs = append(svcs, svc)
-	}
-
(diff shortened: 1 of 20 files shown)


brokerapi/brokers/config/broker_config.go
```diff
@@ -0,0 +1,105 @@
...
+	"os"
+)
+
+type BrokerConfig struct {
+	Catalog    map[string]models.Service
+	HttpConfig *jwt.Config
+	ProjectId  string
+}
+
+func NewBrokerConfig() (*BrokerConfig, error) {
```
johnsonj: suggest: NewBrokerConfigFromEnv

brokerapi/brokers/stackdriver_trace/broker.go
```diff
@@ -56,7 +48,7 @@ func (b *StackdriverTraceBroker) Bind(instanceID, bindingID string, details mode
 	details.Parameters["role"] = "cloudtrace.agent"
 
 	// Create account
-	newBinding, err := b.ServiceAccountManager.CreateAccountInGoogle(instanceID, bindingID, details, models.ServiceInstanceDetails{})
+	newBinding, err := b.AccountManager.CreateAccountInGoogle(instanceID, bindingID, details, models.ServiceInstanceDetails{})
```
johnsonj: now that it's AccountManager there's nothing to indicate ServiceAccount. Google account is pretty generic. Suggest renaming method to something like CreateServiceAccount

ccemeraldeyes: Why do you want it to indicate service account?

johnsonj: So it indicates what it's actually making. AccountInGoogle doesn't have a ton of meaning. What kind of account? google what? a gmail?

brokerapi/brokers/config/broker_config.go
```diff
@@ -0,0 +1,105 @@
...
+	"fmt"
+	"gcp-service-broker/brokerapi/brokers/models"
+	"gcp-service-broker/utils"
+	"golang.org/x/oauth2/jwt"
+	"gopkg.in/validator.v2"
+	"os"
+)
+
+type BrokerConfig struct {
+	Catalog    map[string]models.Service
```
mkjelland: Do you need to store this as a map here? If not, you could store it as a list and then you won't need a public ServiceList method.

brokerapi/brokers/stackdriver_trace/broker.go
ccemeraldeyes: I guess my point is that these brokers are meant to be pretty generic. It shouldn't really matter what kind of account. And the kind of account *is* indicated on creation of the broker. And to have a consistent account manager interface, a generic name is needed anyway.

johnsonj: well they're generic in that they are provisioning a multitude of services but aren't they always GCP services and GCP service accounts? is that ever going to be different?

brokerapi/brokers/config/broker_config.go
```diff
@@ -0,0 +1,105 @@
...
+
+func NewBrokerConfigFromEnv() (*BrokerConfig, error) {
+	var err error
+	bc := BrokerConfig{}
+	creds, err := bc.GetCredentialsFromEnv()
+	if err != nil {
+		return &BrokerConfig{}, err
+	}
+	bc.ProjectId = creds.ProjectId
+	conf, err := utils.GetAuthedConfig()
```
mkjelland: Can GetAuthedConfig be moved out of utils and into this file? This file has a more specific purpose than utils and contains other similar methods.

brokerapi/brokers/stackdriver_trace/broker.go
ccemeraldeyes: CloudSQL ruins everything, yes.

johnsonj: ok, what about CreateAccount/Create/CreateCredentials?

'InGoogle' doesn't sound like it's creating a sql username/password

ccemeraldeyes: sold.

johnsonj: thank you and sorry for the super nit of method names past :)

brokerapi/brokers/config/broker_config.go
ccemeraldeyes: it's used in the db migrations code too so I'm going to leave it.

brokerapi/brokers/config/broker_config.go
ccemeraldeyes: I'm now also storing it as a map on the broker, which is what I meant to do ;)

PR #179, opened 2017-11-10
Remove wait for ssl certificate request to finish

mkjelland:
Waiting for the ssl certificate request to complete makes the bind operation for mysql 2nd generation instances consistently take longer than 60 seconds, which is the default timeout. I'm removing the wait for now and adding a longer sleep to the integration tests so they will still pass.

The ssl certificates are returned before the bind operation finishes, but the instance needs to be updated to use them, which takes 40-60 seconds. The request will continue to run in the background.

### DIFF
--- a/brokerapi/brokers/account_managers/sql_account_manager.go
+++ b/brokerapi/brokers/account_managers/sql_account_manager.go
@@ -74,11 +74,6 @@ func (sam *SqlAccountManager) CreateAccountInGoogle(instanceID string, bindingID
 	if err != nil {
 		return models.ServiceBindingCredentials{}, fmt.Errorf("Error creating ssl certs: %s", err)
 	}
-	certInsertOperation := newCert.Operation
-	err = sam.pollOperationUntilDone(certInsertOperation, sam.ProjectId)
-	if err != nil {
-		return models.ServiceBindingCredentials{}, fmt.Errorf("Error encountered while polling until operation id %s completes: %s", op.Name, err)
-	}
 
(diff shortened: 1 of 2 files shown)


brokerapi/brokers/account_managers/sql_account_manager.go
```diff
@@ -74,11 +74,6 @@ func (sam *SqlAccountManager) CreateAccountInGoogle(instanceID string, bindingID
 	if err != nil {
 		return models.ServiceBindingCredentials{}, fmt.Errorf("Error creating ssl certs: %s", err)
 	}
-	certInsertOperation := newCert.Operation
```
johnsonj: totally agree with this removal unless there's a reason outside of test code.

integration_tests/async_integration_test.go
```diff
@@ -245,6 +245,9 @@ var _ = Describe("AsyncIntegrationTests", func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(credsMap["uri"]).To(ContainSubstring("postgres"))
 
+			// The bind oepration finishes before the instance has finished updating
+			time.Sleep(60 * time.Second)
```
johnsonj: I know it's way more work but did you consider fetching the Operation and polling? This [API](https://cloud.google.com/resource-manager/reference/rest/v1/operations/get) allows you to find a pending operation.

tests relying on time.Sleep() are likely to break

mkjelland: That's a good idea, thanks!

PR #180, opened 2017-11-10
Updating version default text for cloud sql

mkjelland:
This better reflects the format of the version that is expected, and shows the different default values for different instance types.


mkjelland:
ignore this one, git didn't know who I was when I committed this!

Issue #135, opened 2017-04-17
modify cloudsql supplied credentials to make use of proxy libraries


ccemeraldeyes:
Currently cloudsql returns a username, pw, and ssl certs for auth. This decision was made because the only known alternative at the time was to install a proxy service on each VM needing to connect to CloudSQL, which seemed like too much overhead. However, [Java and Go have proxy libraries](https://cloud.google.com/sql/docs/mysql/external-connection-methods?hl=en_US) that can be used with a service account and connection details (instance region in addition to what's provided now) instead.

ccemeraldeyes:
Implementation options:

* always create both sets of creds and return both
* accept a custom parameter (e.g. `cred_type`)
* ???

mgoddard-pivotal:
To me, it's probably okay to just have `VCAP_SERVICES` contain all of it, so you don't need the `cred_type` parameter.  Folks can just ignore whichever elements they don't need.

mgoddard-pivotal:
Here's a simple Spring Boot app which might help illustrate what a CloudSQL binding ought to provide via its `VCAP_SERVICES` (note here the need to also bind to Storage to get the `PrivateKeyData` needed for using the default credentials, and also how the `instanceConnectionName` must be constructed from three chunks of data, one of which isn't yet available via any binding):

  https://github.com/mgoddard-pivotal/gcp-cloudsql-boot

mkjelland:
@mgoddard-pivotal that sample app is really helpful, thank you! I'm looking into this now.

PR #182, opened 2017-11-14
Create service account for binds to cloud sql instances

mkjelland:
For #135

Service accounts will be created when binding to a cloud sql instance. The fields necessary to use the proxying libraries will be returned. These fields are project id, region and private key data.

ssl certs will be generated optionally, based on the generate_ssl_certs parameter to bind.

### DIFF
--- a/brokerapi/brokers/account_managers/sql_account_manager.go
+++ b/brokerapi/brokers/account_managers/sql_account_manager.go
@@ -40,9 +40,14 @@ func (sam *SqlAccountManager) CreateAccountInGoogle(instanceID string, bindingID
 	var err error
 	username, usernameOk := details.Parameters["username"].(string)
 	password, passwordOk := details.Parameters["password"].(string)
+	credType, sslCertsOk := details.Parameters["credential_type"].(string)
 
 	if !passwordOk || !usernameOk {
 		return models.ServiceBindingCredentials{}, errors.New("Error binding, missing parameters. Required parameters are username and password")
+	}
+
+	if !sslCertsOk {
+		credType = "service_account"
 	}
 
 	// create username, pw with grants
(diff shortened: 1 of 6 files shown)


brokerapi/brokers/account_managers/sql_account_manager.go
```diff
@@ -115,25 +127,27 @@ func (sam *SqlAccountManager) DeleteAccountFromGoogle(binding models.ServiceBind
 		return fmt.Errorf("Error creating CloudSQL client: %s", err)
 	}
 
-	op, err := sqlService.SslCerts.Delete(sam.ProjectId, instance.Name, sqlCreds.Sha1Fingerprint).Do()
-	if err != nil {
-		return fmt.Errorf("Error deleting ssl cert: %s", err)
-	}
+	if sqlCreds.CaCert != "" {
```
ccemeraldeyes: I'd at least add a comment here about why. Or, this could be made clearer by retrieving the BindParameters and checking the generate_ssl_certs field.

brokerapi/brokers/account_managers/sql_account_manager.go
```diff
@@ -40,11 +40,16 @@ func (sam *SqlAccountManager) CreateAccountInGoogle(instanceID string, bindingID
 	var err error
 	username, usernameOk := details.Parameters["username"].(string)
 	password, passwordOk := details.Parameters["password"].(string)
+	generateCerts, sslCertsOk := details.Parameters["generate_ssl_certs"].(string)
```
johnsonj: suggest moving it directly about the `if !sslCertsOk`

brokerapi/brokers/account_managers/sql_account_manager.go
```diff
@@ -66,22 +71,29 @@ func (sam *SqlAccountManager) CreateAccountInGoogle(instanceID string, bindingID
...
+		// create ssl certs
+		certname := bindingID[:10] + "cert"
+		newCert, err := sqlService.SslCerts.Insert(sam.ProjectId, instance.Name, &googlecloudsql.SslCertsInsertRequest{
+			CommonName: certname,
+		}).Do()
+		if err != nil {
+			return models.ServiceBindingCredentials{}, fmt.Errorf("Error creating ssl certs: %s", err)
+		}
+
+		creds = SqlAccountInfo{
```
johnsonj: nit: either mutate creds or don't set it unconditionally above

brokerapi/brokers/account_managers/sql_account_manager.go
```diff
@@ -115,25 +127,27 @@ func (sam *SqlAccountManager) DeleteAccountFromGoogle(binding models.ServiceBind
 		return fmt.Errorf("Error creating CloudSQL client: %s", err)
 	}
 
-	op, err := sqlService.SslCerts.Delete(sam.ProjectId, instance.Name, sqlCreds.Sha1Fingerprint).Do()
-	if err != nil {
-		return fmt.Errorf("Error deleting ssl cert: %s", err)
-	}
+	if sqlCreds.CaCert != "" {
+		certOp, err := sqlService.SslCerts.Delete(sam.ProjectId, instance.Name, sqlCreds.Sha1Fingerprint).Do()
```
johnsonj: any risk this times out?

mkjelland: It completes quickly and then we have to wait for the operation to complete below. So I don't think it would timeout.

Issue #172, opened 2017-09-20
Ability to configure AclEntry for CloudSQL Broker


amcculley:
https://github.com/GoogleCloudPlatform/gcp-service-broker/blob/master/brokerapi/brokers/cloudsql/broker.go

Line: 145

Assuming we want to restrict access, where would be the appropriate place to configure this?  Env Var, Plan, ?

ccemeraldeyes:
I'd probably implement this as a parameter to provision

PR #183, opened 2017-11-14
Add authorized networks field to plan details for cloud sql

mkjelland:
For #172

`authorized_networks` is a comma separated list of authorized networks. If no networks are desired, the field should be set to "none".

The field is required in the tile, so if people are updating the service broker, they will be required to fill in a value for this in their plans. I don't believe previously created instances will be updated to respect this new field.


mkjelland:
I'm going to close this one and open another once I confirm that the customer is ok with making this a parameter instead of a plan detail. Thanks for the feedback @ccemeraldeyes

PR #185, opened 2017-12-05
tile: fix spanner node description

### LINKED ISSUE #184: tile: Spanner Number of Nodes Description is wrong
today: "Between 3 and 30 (see https://cloud.google.com/bigtable/pricing for more information)"

johnsonj:
fixes #184

### DIFF
--- a/tile.yml
+++ b/tile.yml
@@ -449,6 +449,6 @@ service_plan_forms:
     - name: num_nodes
       label: Number of Nodes
       type: string
-      description: "Between 3 and 30 (see https://cloud.google.com/bigtable/pricing for more information)"
+      description: "A minimum of 3 nodes is recommended for production environments. (see https://cloud.google.com/spanner/pricing for more information)"
       configurable: true
-      default: '3'
+      default: '1'


tile.yml
```diff
@@ -449,6 +449,6 @@ service_plan_forms:
     - name: num_nodes
       label: Number of Nodes
       type: string
-      description: "Between 3 and 30 (see https://cloud.google.com/bigtable/pricing for more information)"
+      description: "A minimum of 3 nodes is recommended for production environments. (see https://cloud.google.com/spanner/pricing for more information)"
       configurable: true
-      default: '3'
\ No newline at end of file
+      default: '1'
```
mkjelland: Can you explain why the default is changing to 1 which the recommendation is to use 3?

johnsonj: 3 is the recommended min for prod but 1 is fine for other use cases. I am operating under the assumption that a user taking spanner to prod is doing proper capacity planning and not relying on the default, but someone exploring spanner would. 

Happy to go either way with it

PR #186, opened 2017-12-12
Add parameter to provision to specify which networks should be author…

mkjelland:
…ized to access cloudsql instances


docs/use.md
```diff
@@ -86,6 +86,7 @@ Notes:
         * `backup_start_time` (defaults to `"06:00"`)
         * `binlog` (defaults to `false` for 1st gen, true for 2nd gen, set to "true" to use)
         * `activation_policy` (defaults to `on demand`)
+        * `authorized_networks` (defaults to none)
```
ccemeraldeyes: I'd clarify what format is expected here

mkjelland: Good idea, thanks!

PR #188, opened 2017-12-15
Update postgres uri to enable using spring cloud connectors

mkjelland:
Adding certs to the connection uri that is returned in VCAP_SERVICES to applications on bind, and updated go test app to use the uri for connections instead of building it's own uri.

Spring cloud connectors would only connect if I added the `jdbc` in front of the url.

### DIFF
--- a/brokerapi/brokers/account_managers/sql_account_manager.go
+++ b/brokerapi/brokers/account_managers/sql_account_manager.go
@@ -198,8 +198,8 @@ func (b *SqlAccountManager) BuildInstanceCredentials(bindRecord models.ServiceBi
 		combinedCreds["uri"] = fmt.Sprintf("mysql://%s:%s@%s/%s?ssl_mode=required",
 			url.QueryEscape(combinedCreds["Username"]), url.QueryEscape(combinedCreds["Password"]), combinedCreds["host"], combinedCreds["database_name"])
 	} else if service_to_name[sid] == models.CloudsqlPostgresName {
-		combinedCreds["uri"] = fmt.Sprintf("postgres://%s/%s?user=%s&password=%s&ssl=true",
-			combinedCreds["host"], combinedCreds["database_name"], url.QueryEscape(combinedCreds["Username"]), url.QueryEscape(combinedCreds["Password"]))
+		combinedCreds["uri"] = fmt.Sprintf("jdbc:postgres://%s:%s@%s/%s?sslmode=require&sslcert=%s&sslkey=%s&sslrootcert=%s",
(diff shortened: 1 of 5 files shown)


examples/go/sb-test/README.md
```diff
@@ -37,7 +37,8 @@ This is not an official Google product.
 
 * `/test-storage` - lists all buckets in the given project
 * `/test-pubsub` - returns the name of the first topic in the project
-* `/test-cloudsql` - tests the connection to the first bound database
+* `/test-cloudsql-mysql` - tests the connection to the first bound mysql database
+* `/test-cloudsql-postgres` - tests the connection to the first bound postgres database
```
mkjelland: This url was just outdated so I updated it. It isn't super related to this change.

brokerapi/brokers/account_managers/sql_account_manager.go
```diff
@@ -198,8 +198,8 @@ func (b *SqlAccountManager) BuildInstanceCredentials(bindRecord models.ServiceBi
 		combinedCreds["uri"] = fmt.Sprintf("mysql://%s:%s@%s/%s?ssl_mode=required",
 			url.QueryEscape(combinedCreds["Username"]), url.QueryEscape(combinedCreds["Password"]), combinedCreds["host"], combinedCreds["database_name"])
 	} else if service_to_name[sid] == models.CloudsqlPostgresName {
-		combinedCreds["uri"] = fmt.Sprintf("postgres://%s/%s?user=%s&password=%s&ssl=true",
-			combinedCreds["host"], combinedCreds["database_name"], url.QueryEscape(combinedCreds["Username"]), url.QueryEscape(combinedCreds["Password"]))
+		combinedCreds["uri"] = fmt.Sprintf("jdbc:postgres://%s:%s@%s/%s?sslmode=require&sslcert=%s&sslkey=%s&sslrootcert=%s",
```
johnsonj: This discussion has probably happened elsewhere- but:

This is a JDBC specific format and we advertise this a a generic URI. Should be exposed differently in the combinedCreds (binding credentials)? For example as 'jdbc_uri'? 

If it needs to be called URI perhaps URI Format needs to be a binding parameter so the user would so something like:
```bash
cf bind-service myapp mydb -c '{"uri_format":"jdbc"}'
```

mkjelland: That's a good idea. It does to need to be called "uri", because if there is a field called "uri", spring cloud connectors will use that. If there isn't a field called that, it'll looks for a field called "jdbcUrl". Maybe the best solution is to include a uri_format parameter to bind, and then exclude the "uri" field if that option is set, and instead pass back "jdbcUrl". What do you think?

johnsonj: Sounds good to me!

mkjelland: Thanks Jeff!

PR #204, opened 2018-04-19
jbdc -> jdbc

josephlewis42:
Typo fix causing bad JDBC URIs to be returned from the CloudSQL broker.


josephlewis42:
@evandbrown this should be the fix to that URL problem. We can ask @mkjelland to review because she was `git blame`d :-)

evandbrown:
Thanks, @josephlewis42!

PR #206, opened 2018-05-14
Mysql jdbc fix

josephlewis42:
@evandbrown I'd like your thoughts about this as a fix for that problem João is facing.

1. I really hate creating a conditional on that `"jdbc:"` that comes from the map.
2. Would it be possible to add an additional key in combinedCreds for a JDBC variant e.g. `"JdbcUri"` and just include it by default in all results? I don't know enough about what's happening downstream to figure out if that would be exposed to a user. That could mitigate inadvertently breaking someone else's environment.
3. This _should_ be a non-breaking change but it might start exposing username/password pairs in application logs if users were only doing basic sanitation.

I also have to figure out how to test/build this, looks like it needs a custom go setup?

Thoughts?

### DIFF
--- a/brokerapi/brokers/account_managers/sql_account_manager.go
+++ b/brokerapi/brokers/account_managers/sql_account_manager.go
@@ -182,13 +182,23 @@ func (b *SqlAccountManager) BuildInstanceCredentials(bindRecord models.ServiceBi
 	}
 
 	sid := instanceRecord.ServiceId
+	sqlModelName := service_to_name[sid]
 
 	combinedCreds := utils.MergeStringMaps(bindDetails, instanceDetails)
+	isJdbc := combinedCreds["UriPrefix"] == "jdbc:"
 
-	if service_to_name[sid] == models.CloudsqlMySQLName {
+	if sqlModelName == models.CloudsqlMySQLName && isJdbc {
+		q := url.Values{}
+		q.Set("ssl_mode", "required")
+		q.Set("user", combinedCreds["Username"])
+		q.Set("password", combinedCreds["Password"])
+
+		combinedCreds["uri"] = fmt.Sprintf("%smysql://%s/%s?%s",
+			combinedCreds["UriPrefix"], combinedCreds["host"], combinedCreds["database_name"], q.Encode())
+	} else if sqlModelName == models.CloudsqlMySQLName {
 		combinedCreds["uri"] = fmt.Sprintf("%smysql://%s:%s@%s/%s?ssl_mode=required",
 			combinedCreds["UriPrefix"], url.QueryEscape(combinedCreds["Username"]), url.QueryEscape(combinedCreds["Password"]), combinedCreds["host"], combinedCreds["database_name"])
-	} else if service_to_name[sid] == models.CloudsqlPostgresName {
+	} else if sqlModelName == models.CloudsqlPostgresName {
 		combinedCreds["uri"] = fmt.Sprintf("%spostgres://%s:%s@%s/%s?sslmode=require&sslcert=%s&sslkey=%s&sslrootcert=%s",
 			combinedCreds["UriPrefix"], url.QueryEscape(combinedCreds["Username"]), url.QueryEscape(combinedCreds["Password"]), combinedCreds["host"], combinedCreds["database_name"], url.QueryEscape(combinedCreds["ClientCert"]), url.QueryEscape(combinedCreds["ClientKey"]), url.QueryEscape(combinedCreds["CaCert"]))
 	} else {


evandbrown:
First, thanks for taking this on, Joseph!

1. I think checking the `UriPrefix` is ok. It's now basically a flag since you discovered that the JDBC and non-JDBC connection strings differ on more than just that prefix. 

2. According to [this doc](https://github.com/GoogleCloudPlatform/gcp-service-broker/blob/master/docs/use.md), if `jdbc_uri_format` is true then the "`uri` field will contain a jdbc formatted uri". Although it's [possible to add](https://docs.cloudfoundry.org/services/binding-credentials.html) another key (like `JdbcUri`), I'm thinking we should acknowledge the `jdbc_uri_format` flag if set and make `uri` contain the JDBC string.

3. Is there a reason this would expose credentials any different than the non-JDBC or Postgres modes do?

evandbrown:
Oh and the `go` setup is a little weird. Here's how I'd do it (you'll want to install `direnv` if you don't already have it):

```
$ mkdir -p gcp-service-broker/src
$ git clone git@github.com:GoogleCloudPlatform/gcp-service-broker.git gcp-service-broker/src/gcp-service-broker
$ echo 'export GOPATH=`pwd`
export PATH=$PATH:$GOPATH/bin' > gcp-service-broker/.envrc
$ cd gcp-service-broker/src/gcp-service-broker
$ direnv allow
$ go build cmd/migrate/migrate.go
```

Oh wow, I just did that and saw that the `migrate` binary is checked in. Yikes. I added that to #207

evandbrown:
Ahh, gotcha on the Spring bit. Since the bound credentials are exposed as env vars and we don't really control what the app/framework/dev does with them if they read them, I'm inclined to say it's OK. It would be interesting to observe how vanilla Spring behaves, though.

LMK if I can help w/ the Go setup beyond my previous comment. Once a test is added (or fixed) for the JDBC stuff, we should be able to merge and send on to integration testing.

joaoandremartins:
Based on [this article](https://spring.io/blog/2015/04/27/binding-to-data-services-with-spring-boot-in-cloud-foundry), it seems that the provided JDBC URL has that weird shape on purpose:

> NOTE: The URL in the VCAP_SERVICES is actually not a “jdbc” scheme, which should be mandatory for JDBC connections. This is, however, the format that Cloud Foundry normally presents it in because it works for nearly every language other than Java. Spring Cloud Connectors or the buildpack auto-reconfiguration, if they are creating a DataSource, will translate it into a jdbc:* URL for you.

It's still not clear to me what, in Cloud Foundry, is reading and parsing this JDBC URL.

Also, [the JDBC URL recommended by the Cloud SQL team](https://cloud.google.com/sql/docs/mysql/connect-external-app) points the JDBC driver to a special tunneling `SocketsFactory`:

> jdbc:mysql://google/%s?cloudSqlInstance=%s&socketFactory=com.google.cloud.sql.mysql.SocketFactory&useSSL=false

If we can't use this in CF and need to use the default socket factory, then it's unclear to me what the value our starters can offer in CF is, and I would prefer to back off and let users configure their JDBC setup the same way as they used to do.

Finally, I'm finding that something in CF is disregarding our provided `DataSourceProperties` (probably what they call "auto-reconfiguration").
For example, using the `spring-cloud-gcp-starter-sql-mysql` and making a query using `JdbcTemplate`, I can verify that `DataSourceProperties` is well configured but the resulting `DataSource` is configured with a `jdbc:mysql://null/null` URL.
This is also probably auto-reconfiguration's doing, which disregards the provided `DataSourceProperties`, and even the `DataSource` provided by Spring Boot, and instead tries to build a `DataSource` from `VCAP_SERVICES`, at which point it might not be able to detect our `google-cloudsql-mysql` configuration there and produces that useless JDBC URL/`DataSource`.

saturnism:
/cc @nebhale

Ben, in this issue, it reads like that the Java Buildpack is reading/parsing the VCAP services configuration, and then applying auto-reconfiguration for Spring apps at deployment time by adding new dependencies and also configuring JDBC URIs, etc.

In Spring Cloud GCP, we have corresponding starters that can already extract VCAP information and construct this URI and many other properties.  However, it looks like that the existing Java buildpack also injected other code to override and/or change this configuration.

I'm trying to see how we can solve this issue, e.g.:
- avoid auto-reconfiguration by CF buildpack
- or, add Spring Cloud GCP starter as part of the buildpack auto-reconfiguration
- or, a combination of both
- or maybe there is something else we are not aware of :)

Thanks!

nebhale:
I agree that this appears to be auto-reconfiguration kicking in an overwriting an existing `DataSource` with one configured to the contents of `VCAP_SERVICES`.  Auto-configuration is pretty thin, wrapping around [Spring Cloud Connectors](https://cloud.spring.io/spring-cloud-connectors/spring-cloud-cloud-foundry-connector.html#_mysql) and if I had to guess the GCP service flags on the detection criteria for that project.

I see two possibilities for solving this and I'd like @scottfrederick (lead of Spring Cloud Connectors) to weigh in with his suggestions.  The first option is for SCC to gain direct knowledge of the GCP services and find some way to avoid them or otherwise configure them properly.  The second option is to extend our "disable auto-reconfiguration" heuristics from just noticing that someone has used SCC to include both SCC and the GCP starter.  To do the latter, we'd need to be able to find some bean type that is GCP-specific in the `ApplicationContext` (maybe some shared credential type?).

I think the advantage of the former is that it could handle this issue without us disabling all auto-reconfiguration in globally, but I'm certainly sensitive to that pushing work onto another team when the "big hammer" solution might be sufficient.

scottfrederick:
Just for reference, there has been some discussion on Spring Cloud Connectors and Google Cloud SQL in [an issue](https://github.com/spring-cloud/spring-cloud-connectors/issues/191) in the SCC Github repo. Looking at the example `VCAP_SERVICES` payload in that issue, SCC must be detecting on the `uri` field with the scheme `mysql://`. 

It wouldn't be hard to create an SCC extension project to create a properly configured `DataSource` as suggested at the end of that issue thread, but that seems to compete with the starters approach in `spring-cloud-gcp`. 

Alternatively, the core SCC library that contains the service type detection logic could be modified to disqualify a Google Cloud SQL service if found in `VCAP_SERVICES`. This would have to be done by looking for GCP-specific values in the `label` field (e.g. `google-cloudsql`) or the `tags` field (e.g. `gcp`, `cloudsql`). We haven't had to do this for other vendor-provided DBs, and it feels a bit hacky to me. Such a change would require a new release of SCC, and a new release of the Java buildpack to pull in the new SCC release. Users with older versions of the JBP would still have the problem.

nebhale:
I think that no matter what the solution is, it'll be a going forward fix but until then, today's ability to disable auto-reconfiguration with `cf set-env <APP> JBP_CONFIG_SPRING_AUTO_RECONFIGURATION '{enabled: false}'` will be the work-around.

Another reason that this might be best fixed in SCC is that putting the disabling logic in the buildpack doesn't help the situation where SCC and the GCP starter both exist together in an application.  Is that likely to be common enough that it tips the scales in one direction or another?

scottfrederick:
> the situation where SCC and the GCP starter both exist together in an application. Is that likely to be common enough that it tips the scales in one direction or another?

I'd say that's a likely combination given that apps often use SCC to configure other connections.

saturnism:
@joaoandremartins can we try and document the workaround `cf set-env <APP> JBP_CONFIG_SPRING_AUTO_RECONFIGURATION '{enabled: false}' ` when using Spring Cloud GCP w/ PCF? That seems to solve all of our issues.

@scottfrederick It would be great to have a `spring-cloud-gcp-connector` of some sort to make this seamless w/ PCF deployments. That would be ideal for the long term. How do we get started? :)

joaoandremartins:
Huzzah! I finally got my CF cluster to start working again, and I verified that `cf set-env <APP> JBP_CONFIG_SPRING_AUTO_RECONFIGURATION '{enabled: false}'` works like a charm.

We'll be documenting this as a solution on our side.

In the meantime, it sounds like some work still needs to be done on the GCP CF CB side, namely ensuring that the current URI gets picked up by whatever is doing the auto reconfiguration on the Java buildpack, so that we don't end up with a `jdbc:mysql://null/null` URL in the `DataSource`.

Not sure if in scope here, but we should also be using the [Google Cloud SQL sockets factory](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory). If we can do that without resorting to our starters, the URL should also be built differently, e.g., `jdbc:mysql://google/...`.

Also FYI that the documentation in https://docs.pivotal.io/partners/gcp-sb/using.html needs to be updated. Specifically, in the SQL bit, `google-cloudsql` service name no longer exists, it's now divided in `google-cloudsql-mysql` and `google-cloudsql-postgres`. The `d4_standard` plan also no longer exists.

evandbrown:
Woohoo!

PR #214, opened 2018-07-13
re-package and fix compile errors

josephlewis42:


josephlewis42:
Last of the work to get Colleen's work merged and building. Master drifted since she started so I had to re-add `SaAccountManager models.AccountManager` back into the SQL manager.

PR #218, opened 2018-07-17
Use Viper rather than `os.GetEnv`

### LINKED ISSUE #158: add envconfig repo

josephlewis42:
Rather than `os.GetEnv()` for every environment call, we use the Viper registry. If the user specifies a `--config` in TOML, YAML, JSON, or .properties formats, it will be read and used.

This will allow us to more easily document what flags we're using, write tests, refactor environment names safely, integrate with CLI flags and opens the way towards using key/value configuration systems like [etcd or consul](https://github.com/spf13/viper) and potentially [Vault](https://github.com/spf13/viper/issues/306) further down the line.

fixes #158


cmd/root.go
```diff
@@ -49,3 +53,22 @@ func Execute() {
...
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	viper.AutomaticEnv()
+}
+
+func initConfig() {
+	if cfgFile != "" {
+		viper.SetConfigFile(cfgFile)
+
+		if err := viper.ReadInConfig(); err != nil {
+			fmt.Println("Can't read config:", err)
```
craigatgoogle: These two lines can be combined into a single: log.Fatalf()

cmd/serve.go
```diff
@@ -65,16 +70,21 @@ func serve() {
...
 	// init api
+	logger.Info("Serving", lager.Data{
+		"port":     viper.GetString("api.port"),
+		"username": username,
+	})
+
 	brokerAPI := brokerapi.New(serviceBroker, logger, credentials)
 	http.Handle("/", brokerAPI)
-	http.ListenAndServe(":"+os.Getenv("PORT"), nil)
+	http.ListenAndServe(":"+viper.GetString("api.port"), nil)
```
craigatgoogle: Instead of retrieving "api.port" twice, consider doing it once and assigning to a variable, similar to username and password.

db_service/setup_db.go
```diff
@@ -27,29 +27,48 @@ import (
 	"code.cloudfoundry.org/lager"
 	"github.com/go-sql-driver/mysql"
 	"github.com/jinzhu/gorm"
+	"github.com/spf13/viper"
 )
 
+const (
+	caCertProp     = "db.ca.cert"
```
craigatgoogle: This approach seems inconsistent with serve.go. I'm not sure what the benefit is of adding a layer of indirection as opposed to using the strings in-line.

josephlewis42: These are used twice, so I wanted to avoid a double hard-coded variable in case one got changed. What do you think about pulling these out into a separate config package?

Something that would hold string constants and set up the legacy mappings? I always feel a little bit iffy about that as a design pattern because I feel like it lets access creep deep into the code which means people start panicing rather than returning errors which I feel is an anti-pattern. But maybe it can't be helped in this case.

Thoughts?

craigatgoogle: I could see the benefit of either approach. My primary appeal is for consistency across the code base. As you're closer to this code, I'll defer to your best judgement :)

PR #220, opened 2018-07-18
added OSB client for automated testing

josephlewis42:
Added an OSB compatible client with JSON outputs so we build true E2E tests from it and test that functionality hasn't changed from whatever OSB version we're on now to 2.13 where we're going:

```
gcp-service-broker client catalog
gcp-service-broker client provision --instanceid abc --serviceid 0000-1111-2222 --planid 12345 --params '{"region":"US"}'
```

This PR also brings in the `pivotal-cf/brokerapi` library which we're going to replace our API implementation with.


pkg/client/client.go
```diff
@@ -0,0 +1,131 @@
...
+			return nil, err
+		}
+	}
+
+	request, err := http.NewRequest(method, url.String(), buffer)
+	if err != nil {
+		return nil, err
+	}
+
+	request.Header.Set("X-Broker-Api-Version", "2.13")
```
craigatgoogle: Would it make sense to promote the version somewhere more prominently located? I could see this getting overlooked during a release as is. Also, if not already, consider adding a note to the documentation mentioning this needs to be modified with each new version.

josephlewis42:
@craigatgoogle I pulled out that constant and added some notes. It may be worth exploring down the line if we want to make the version configurable via the CLI.

I'm planning on combining the client with the structs we're going to use internally to generate documentation/tile.yml/manifest.yml to create test-cases from embedded examples (executable documentation) so it'll validate the docs, our client, and the broker work all in one go.

PR #223, opened 2018-07-30
Executable Docs

josephlewis42:
This commit brings four things to the service-broker:

 * Docs get descriptions and notes from MagicModules where they were lacking before.
 * Docs, code, tests, and *.yml files are now derived from a single source of truth (the source code).
 * Examples of how to use provision/bind for different scenarios--these examples get generated into documentation for users, and are executable so they can be used to test a new deployment _and_ as integration tests. We know the docs are always good and because we use an OSB compatible HTTP client we also get a true E2E test.
 * New default plans for all services so the broker will work out of the box. CloudSQL comes pre-loaded with plans for all node sizes, Spanner gets a sandbox and production version, and BigTable gets HDD and SSD plans.

You can run the full integration tests like this:

    ./gcp-service-broker client run-examples

And generate the documentation like this:

	./gcp-service-broker generate use

Next up is getting Concourse to run these new integration tests on-demand.


docs/use.md
```diff
@@ -1,235 +1,852 @@
...
-        ```
-        "credentials": {
-             "Email": "pcf-binding-abc123@projectid.iam.gserviceaccount.com",
-             "Name": "pcf-binding-abc123",
-             "PrivateKeyData": "redacted",
-             "ProjectId": "projectid",
-             "UniqueId": "12345",
-        }
-        ```
+
```
josephlewis42: Auto-generated, see rendered docs here: https://github.com/GoogleCloudPlatform/gcp-service-broker/blob/3d471b17a808b08db16c7c8681c937600c555a0e/docs/use.md

pkg/broker/registry.go
```diff
@@ -0,0 +1,133 @@
...
+	var out []*BrokerService
+
+	for _, svc := range brokerRegistry {
+		out = append(out, svc)
+	}
+
+	return out
+}
+
+type BrokerService struct {
```
josephlewis42: I'm not a huge fan of this structure where the clients register something that has to be set up with an init() once done, I'd prefer two separate structs but wasn't 100% sure how to get a good solution for two different back-ends (Terraform and native) until I have something to work with and we have the new REST API in.

pkg/broker/variables.go
```diff
@@ -0,0 +1,37 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package broker
+
+const (
+	JsonTypeString  JsonType = "string"
```
josephlewis42: Unfortunately, all existing code _requires_ values to be strings which it then parses. In the future we can choose to use appropriate types.

TESTING.md
```diff
@@ -321,3 +54,14 @@ api:
   password: pass
   port: 8000
 ```
+
+
+## Useful commands
+
+Create unbind commands for all bindings:
+
+  ./gcp-service-broker show bindings | jq --raw-output '.[] | "gcp-service-broker client unbind --bindingid \(.BindingId) --instanceid \(.ServiceInstanceId) --planid \(.PlanId) --serviceid \(.ServiceId)"'
```
craigatgoogle: Would it make sense to bypass the copy/paste step for the user and just execute the commands with xargs?

brokerapi/brokers/account_managers/sql_account_manager.go
```diff
@@ -69,9 +70,9 @@ func (sam *SqlAccountManager) CreateCredentials(instanceID string, bindingID str
 
 	var creds SqlAccountInfo
 	// create ssl certs
-        certname := bindingID + "cert"
+	certname := bindingID + "cert"
 	if len(bindingID) >= 10 {
```
craigatgoogle: It feels like this hard-coded length and subsequent slice range could benefit for a constant.

TESTING.md
josephlewis42: I like that idea, so far I've had to slightly tweak the outputs so they use the same configuration options as the current command (`--config ../minimal.yml`).

What do you think about a `./gcp-service-broker client unbind-all` command? I keep wavering about it potentially being too dangerous but it would be really handy.

brokerapi/brokers/account_managers/sql_account_manager.go
josephlewis42: I'll add a ticket for these because they're all over the place, service account lengths, project name lengths, etc. At least the service account is labeled with a BUG tag, but I don't know if it was ever updated or planned to be.

brokerapi/brokers/bigquery/definition.go
```diff
@@ -0,0 +1,82 @@
...
+	accountmanagers "github.com/GoogleCloudPlatform/gcp-service-broker/brokerapi/brokers/account_managers"
+	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/broker"
+)
+
+func init() {
+	bs := &broker.BrokerService{
+		Name: "google-bigquery",
+		DefaultServiceDefinition: `{
+        "id": "f80c0a3e-bd4d-4809-a900-b4e33a6450f1",
+        "description": "A fast, economical and fully managed data warehouse for large-scale data analytics",
```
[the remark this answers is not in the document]
josephlewis42: I like where you're going with this, there's also an API "requirement" that we list prices with our plans that would benefit from real-time info.

Maybe starting with the Terraform services, I can find a way for us to run the queries for each field to get descriptions too. I can talk to the MM folks to see what they've pioneered.

brokerapi/brokers/storage/definition.go
```diff
@@ -0,0 +1,101 @@
...
+				Details:   "Name of the bucket this binding is for",
+			},
+		),
+
+		Examples: []broker.ServiceExample{
+			broker.ServiceExample{
+				Name:            "Basic Configuration",
+				Description:     "Create a nearline bucket with a service account that can create/read/delete the objects in it.",
+				PlanId:          "a42c1182-d1a0-4d40-82c1-28220518b360",
+				ProvisionParams: map[string]interface{}{"location": "us"},
```
craigatgoogle: Isn't a bucket name required for provisioning as well?

cmd/generate.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/generator"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	rootCmd.AddCommand(generateCmd)
+	generateCmd.AddCommand(generateUseCmd)
+}
+
+var generateCmd = &cobra.Command{
```
craigatgoogle: I wonder if it might be more concise to define these in-line inside of the init() function rather than outside in a variable which otherwise goes unused. Same question for similar files throughout this CL.

pkg/broker/example.go
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package broker
+
+// ServiceExample holds example configurations for a service that _should_
+// work.
+type ServiceExample struct {
+	// Name is a human-readable name of the example
```
craigatgoogle: Nit, some of your single line comments end with a period and some don't.

pkg/broker/registry.go
```diff
@@ -0,0 +1,133 @@
...
+	"log"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/gcp-service-broker/brokerapi/brokers/models"
+	"github.com/spf13/viper"
+)
+
+var brokerRegistry = make(map[string]*BrokerService)
+
+func Register(service *BrokerService) {
```
craigatgoogle: Missing godoc on this exported function.

pkg/broker/registry.go
craigatgoogle: I hear what you're saying. If it helps, the init() style setup you reference seems to be idiomatic with other Go code I've encountered.

craigatgoogle: Oh, I think I understand what you mean a little better. One option is to define a struct abstraction for DefaultServiceDefinition, and then have separate Marshal functions for the different back-ends. It would add some more structure and type safety and make things a little less brittle overall.

pkg/broker/registry.go
```diff
@@ -0,0 +1,133 @@
...
+func (svc *BrokerService) IsEnabled() bool {
+	return viper.GetBool(svc.enabledProperty)
+}
+
+func (svc *BrokerService) CatalogEntry() models.Service {
+	return svc.serviceDefinition
+}
+
+func (svc *BrokerService) GetPlanById(planId string) *models.ServicePlan {
+	for _, plan := range svc.CatalogEntry().Plans {
```
craigatgoogle: Are there other parts of the system which depend upon plans being a list? If not, it might make sense to convert it to a map for this use-case.

pkg/client/example-runner.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+func RunExamplesForService(client *Client, serviceName string) error {
+	rand.Seed(time.Now().UTC().UnixNano())
+
+	services := broker.GetAllServices()
+
+	for _, service := range services {
+
+		if serviceName != "" && serviceName != service.Name {
```
craigatgoogle: serviceName != "" seems redundant assuming all of the entries withing services have a populated name field.

craigatgoogle: Unless serviceName = "" implies you want all services. It might make sense to document that assumption in the godoc for this function.

pkg/client/example-runner.go
```diff
@@ -0,0 +1,287 @@
...
+	"fmt"
+	"log"
+	"math/rand"
+	"time"
+
+	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/broker"
+	"github.com/pivotal-cf/brokerapi"
+)
+
+func RunExamples(client *Client) error {
```
craigatgoogle: The exported functions in this file are missing godoc.

pkg/client/example-runner.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	// Check that the binding response has the same fields as expected
+	var binding brokerapi.Binding
+	err = json.Unmarshal(bindResponse, &binding)
+	if err != nil {
+		return err
+	}
+
+	credentialsEntry := binding.Credentials.(map[string]interface{})
```
craigatgoogle: I understand that this isn't our API, but man is this brittle. </rant>

pkg/client/example-runner.go
```diff
@@ -0,0 +1,287 @@
...
+		return resp.Error
+	}
+
+	switch resp.StatusCode {
+	case 201:
+		return nil
+	case 202:
+		return ee.pollUntilFinished()
+	default:
+		return errors.New(fmt.Sprintf("Unexpected response code %d", resp.StatusCode))
```
craigatgoogle: This can be made more concise with fmt.Errorf(...)

brokerapi/brokers/storage/definition.go
josephlewis42: A bucket name will be auto-generated if not provided. I decided to go that route because with buckets having globally unique names I couldn't guarantee that we'd hold the name for this bucket between tests (especially if someone else tried to run the examples).

pkg/client/example-runner.go
josephlewis42: Thanks, this reminded me to open #226!

pkg/client/example-runner.go
josephlewis42: ++

brokerapi/brokers/storage/definition.go
craigatgoogle: Makes sense.

cmd/generate.go
craigatgoogle: Discussed this offline with jlewis, we agreed to push this change to a separate PR.

PR #225, opened 2018-07-31
Use New Registry

josephlewis42:
Make the broker reference services from the new registry.

Now custom configuration is pulled in exactly ONE place. This allows #51 to be fixed as a side-effect because now a user can set the following properties to disable services:

```
service:
  google-cloudsql-mysql:
    enabled: false
  google-cloudsql-postgres:
    enabled: false
```

and/or the following environment variables:

```
GSB_SERVICE_GOOGLE_PUBSUB_ENABLED=false
```

...or through the new tile form.

### DIFF
--- a/tile.yml
+++ b/tile.yml
@@ -71,7 +71,120 @@ forms:
     type: text
     label: Client key
     optional: true
-
+- name: enable_disable
+  label: Enable Services
+  description: Enable or Disable Services
+  properties:
+    - name: gsb_service_google_bigquery_enabled
+      label: Let end-users create plans for google_bigquery
+      type: dropdown_select
+      configurable: true
+      default: 'true'
+      options:
+        - name: 'true'
+          label: 'Enable'
+        - name: 'false'
+          label: 'Disable'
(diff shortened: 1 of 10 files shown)


tile.yml
```diff
@@ -71,7 +71,120 @@ forms:
     type: text
     label: Client key
     optional: true
-
+- name: enable_disable
```
craigatgoogle: If it's not already, is it possible to have this file generated using the data in the various brokers/{service_name}/definition.go files? Seems like it would save potential manual maintenance in the future.

josephlewis42: This is something I'd like to look at. I think we might be able to generate the service plan forms from the same service definitions we already have coded in the registry (adding one more set per-service for the user-configurable stuff).

These forms I'd really like to be able to do in a similar way, but also generate the documentation for `docs/customization.md` because not everyone is using this as an installed tile and those people need docs.

I think my general approach is going to be to write a separate config package that wraps viper and can keep track of/generate docs at the same time.

PR #227, opened 2018-08-01
Cleanup the CLI commands

josephlewis42:
Refactoring the CLI commands into init() functions so they doesn't litter the namespace with single-use variables.

There are two intentional functionality changes in here:

* You can now dump the operations table too in the database dumping sub-command.
* The API client now uses `log.Fatalf` rather than returning an error, to Cobra returning an error means a configuration issue so it prints out the help of the command too which caused somewhat confusing errors.


cmd/client.go
```diff
@@ -159,10 +149,18 @@ func newClientCommand(use, short string, run func(*client.Client) *client.Broker
 		RunE: func(cmd *cobra.Command, args []string) error {
```
craigatgoogle: This should be changed to Run:

cmd/client.go
```diff
@@ -174,13 +172,3 @@ func createClient() (*client.Client, error) {
 
 	return client.New(user, pass, "localhost", port)
 }
-
-func printJsonResults(results interface{}) error {
```
craigatgoogle: This JSON formatting functionality seems to be repeated throughout the PR. Any reason in particular to not have it live in this helper function?

josephlewis42: Nice catch, I pulled it out into the util package.

utils/utils.go
```diff
@@ -61,3 +62,15 @@ func MergeStringMaps(map1 map[string]string, map2 map[string]string) map[string]
 	}
 	return combined
 }
+
+// PrettyPrintOrExit writes a JSON serialized version of the content to stdout.
+// If a failure occurs during marshaling, the error is logged along with a
+// formatted version of the object and the program exits with a failure status.
+func PrettyPrintOrExit(content interface{}) {
```
craigatgoogle: Perhaps make whether or not to exit an argument, that way this method will still work for the other use-cases throughout this PR which need this behavior but with a returned error.

craigatgoogle: Or else, have a separate method which returns an error, and have both use a private helper.

craigatgoogle: The second way sounds cleaner the more I think about it.

josephlewis42: I like that, it'll also go cleanly with the way the rest of this software is structured.

PR #230, opened 2018-08-02
OSB2 API

josephlewis42:
This PR moves us to using the CF sanctioned OSB Provider library and structs so upgrading to match future updates to the spec won't be so painful and we can rely on their well tested library to use the proper field names, response codes, return types, etc.

Ideally, I'd like to delete most of `integration_test` and `async_integration_test` now that we have our executable end to end tests. Many of the tests are duplicates, rely on magic numbers or implementation details e.g. expecting exactly 11 services to be in the catalog rather than checking `len(catalog) == len(service_name_list)` or querying the database with custom SQL.

In the 4.1.0 release I'm planning on getting us to use proper type casting and allow user-defined defaults in prep of adding Terraform support which will clean up the marshaling/unmarshaling I stuck in the methods due to the upstream library using `json.RawMessage` rather than `interface[string]interface{}`s you see now. `gcp_service_broker.go` will also get refactored once Terraform support rolls in to work with multiple backends.


brokerapi/brokers/account_managers/service_account_manager.go
```diff
@@ -43,10 +44,15 @@ type ServiceAccountManager struct {
...
 // creates a new service account for the given binding id with the role listed in details.Parameters["role"]
-func (sam *ServiceAccountManager) CreateCredentials(instanceID string, bindingID string, details models.BindDetails, instance models.ServiceInstanceDetails) (models.ServiceBindingCredentials, error) {
+func (sam *ServiceAccountManager) CreateCredentials(instanceID string, bindingID string, details brokerapi.BindDetails, instance models.ServiceInstanceDetails) (models.ServiceBindingCredentials, error) {
 	client := sam.HttpConfig.Client(context.Background())
 
-	role, ok := details.Parameters["role"].(string)
+	bindParameters := map[string]interface{}{}
+	if err := json.Unmarshal(details.RawParameters, &bindParameters); err != nil {
+		return models.ServiceBindingCredentials{}, err
+	}
```
josephlewis42: This JSON stuff will be replaced later with a proper programming-language style resolution table where variables will be resolved in the order `instance details -> bind params -> operator provided defaults -> hard-coded defaults`

brokerapi/brokers/brokers_test.go
```diff
@@ -178,84 +174,54 @@ var _ = Describe("Brokers", func() {
...
+			serviceList, err := gcpBroker.Services(context.Background())
+			Expect(err).ToNot(HaveOccurred())
 			for _, s := range serviceList {
 				if s.ID == serviceNameToId[models.StackdriverDebuggerName] {
 					Expect(len(s.Plans)).To(Equal(1))
 				}
 			}
 		})
 
-		It("should error if plan ids are not supplied", func() {
```
josephlewis42: This resolution is no longer handled at this point in the code. It's caught on startup and will fail early rather than later (same with the below portion of the SQL)

brokerapi/brokers/account_managers/sql_account_manager.go
```diff
@@ -39,10 +40,16 @@ type SqlAccountManager struct {
...
+		return models.ServiceBindingCredentials{}, err
+	}
+
 	var err error
-	username, usernameOk := details.Parameters["username"].(string)
-	password, passwordOk := details.Parameters["password"].(string)
+	username, usernameOk := bindParameters["username"].(string)
+	password, passwordOk := bindParameters["password"].(string)
 
 	if !passwordOk || !usernameOk {
```
craigatgoogle: Even though it's less concise, it might make sense to split these into two separate error messages for clarity to save users some time while debugging.

brokerapi/brokers/cloudsql/broker.go
```diff
@@ -461,10 +469,15 @@ func (b *CloudSQLBroker) Bind(instanceID, bindingID string, details models.BindD
...
-	jdbcUriFormat, jdbcUriFormatOk := details.Parameters["jdbc_uri_format"].(string)
+	params := make(map[string]interface{})
+	if err := json.Unmarshal(details.RawParameters, &params); err != nil {
+		return models.ServiceBindingCredentials{}, fmt.Errorf("Error unmarshalling parameters: %s", err)
+	}
+
+	jdbcUriFormat, jdbcUriFormatOk := params["jdbc_uri_format"].(string)
 	credsJSON["UriPrefix"] = ""
 	if jdbcUriFormatOk && jdbcUriFormat == "true" {
-		credsJSON["UriPrefix"] = "jbdc:"
```
craigatgoogle: Ouch, that's a sneaky bug :)

brokerapi/brokers/bigquery/broker.go
```diff
@@ -105,7 +107,7 @@ func (b *BigQueryBroker) Deprovision(instanceID string, details models.Deprovisi
 
 	dataset := models.ServiceInstanceDetails{}
 	if err = db_service.DbConnection.Where("ID = ?", instanceID).First(&dataset).Error; err != nil {
-		return models.ErrInstanceDoesNotExist
+		return brokerapi.ErrInstanceDoesNotExist
```
evandbrown: I wonder if there are any callers that do a type assertion on the old type?

brokerapi/brokers/bigtable/broker.go
```diff
@@ -139,7 +141,7 @@ func (b *BigTableBroker) Deprovision(instanceID string, details models.Deprovisi
 
 	instance := models.ServiceInstanceDetails{}
 	if err = db_service.DbConnection.Where("ID = ?", instanceID).First(&instance).Error; err != nil {
-		return models.ErrInstanceDoesNotExist
+		return brokerapi.ErrInstanceDoesNotExist
```
evandbrown: Same question as previous about callers doing a type assertion on the old error type...

brokerapi/brokers/bigquery/broker.go
josephlewis42: Behind the scenes the API translates these into HTTP response codes and error messages so we should be alright. The strings should be the same as well because we originally copy/pasted their code and they've only refactored that section only slightly since.

Issue #231, opened 2018-08-02
Generate Customization Documentation and Forms


josephlewis42:
Create generators for and update:

 - [x] Enable/Disable services form
 - [x] Custom Service Plan Forms
 - [x] `customizaiton.md` - should include documentation for enable/disable environment variables, custom service plans, and environment variables used for database setup

PR #233, opened 2018-08-06
add custom plan parsing

### LINKED ISSUE #157: Break preconfigured plans into a plan per service
This will help with finer grained service controls - no need to provide a plan for a service you're not going to use. This is especially helpful when the broker is exposed as a tile.

josephlewis42:
Helps fix #157, #155 

We can now read plans from the environment/viper variable for a given broker and parse append them to the main catalog's plans. Plus, tests for the above.

This leads us down the path toward fixing #231

### DIFF
--- a/pkg/broker/broker_test.go
+++ b/pkg/broker/broker_test.go
@@ -16,6 +16,7 @@ package broker
 
 import (
 	"fmt"
+	"testing"
 
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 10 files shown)


pkg/broker/registry.go
```diff
@@ -77,7 +86,9 @@ func MapServiceIdToName() map[string]string {
 	out := map[string]string{}
 
 	for _, svc := range brokerRegistry {
-		out[svc.CatalogEntry().ID] = svc.Name
```
evandbrown: Does `svc.CatalogEntry()` not return an error here because this is an older version of the library?

pkg/broker/broker_test.go
```diff
@@ -64,3 +65,176 @@ func ExampleBrokerService_IsEnabled() {
 	// Output: true
 	// false
 }
+
+func ExampleBrokerService_ServiceDefinition() {
```
evandbrown: Are these `Example...` funcs to demonstrate usage?

josephlewis42: Yes, they're the executable examples that show up in godoc but also count towards unit test coverage. I wanted to document the common functions as examples to show how they interact with Viper.

pkg/broker/registry.go
josephlewis42: This is me refactoring CatalogEntry so it re-parses the catalog entry on every call to be more idiomatic viper-y. It gets rid of all that pre-processing work done when you register a service at some extra CPU expense, but also readies us for no downtime config refreshes with only SIGHUP.

pkg/broker/broker_test.go
```diff
@@ -64,3 +65,176 @@ func ExampleBrokerService_IsEnabled() {
...
+
+	// Override
+	viper.Set(service.DefinitionProperty(), `{"id":"override-id"}`)
+	defn, err = service.ServiceDefinition()
+	fmt.Printf("%q %v\n", defn.ID, err)
+
+	// Bad Value
+	viper.Set(service.DefinitionProperty(), "nil")
+	_, err = service.ServiceDefinition()
+	fmt.Printf("%v\n", err == nil)
```
craigatgoogle: It could make sense to print the error so that the target audience gets an idea of what it might look like. Could be helpful for them  to be better able to identify this kind of error later on while debugging.

craigatgoogle: Prefixed with something like, "Example error: %v\n"

pkg/broker/broker_test.go
```diff
@@ -64,3 +65,176 @@ func ExampleBrokerService_IsEnabled() {
...
+	service := BrokerService{
+		Name: "left-handed-smoke-sifter",
+		DefaultServiceDefinition: `{"id":"abcd-efgh-ijkl", "plans": [{"id": "builtin-plan", "name": "Builtin!"}]}`,
+	}
+
+	viper.Set(service.UserDefinedPlansProperty(), `[{"id":"custom-plan", "name": "Custom!"}]`)
+	defer viper.Set(service.UserDefinedPlansProperty(), nil)
+
+	plan, err := service.GetPlanById("builtin-plan")
+	fmt.Printf("%q %v\n", plan.Name, err)
```
craigatgoogle: It could be helpful for reading the output to prefix each Printf with the use-case being demonstrated, eg: fmt.Printf("builtin-plan: %q %v\n")

pkg/broker/broker_test.go
```diff
@@ -64,3 +65,176 @@ func ExampleBrokerService_IsEnabled() {
...
+		plans, err := service.UserDefinedPlans()
+
+		// Check errors
+		hasErr := err != nil
+		if hasErr != tc.ExpectError {
+			t.Errorf("%s) Expected Error? %v, got error: %v", tn, tc.ExpectError, err)
+		}
+
+		// Check IDs
+		if len(plans) != tc.PlanCount {
```
craigatgoogle: In addition to verifying the correct number of plans, it could be helpful to also verify the plan IDs were correctly propagated. Perhaps adding expectedIDs to the test case struct.

pkg/broker/broker_test.go
```diff
@@ -64,3 +65,176 @@ func ExampleBrokerService_IsEnabled() {
...
+		viper.Set(service.DefinitionProperty(), tc.UserDefinition)
+		viper.Set(service.UserDefinedPlansProperty(), tc.UserPlans)
+
+		srvc, err := service.CatalogEntry()
+		hasErr := err != nil
+		if hasErr != tc.ExpectError {
+			t.Errorf("%s) Expected Error? %v, got error: %v", tn, tc.ExpectError, err)
+		}
+
+		if err == nil && len(srvc.Plans) != tc.PlanCount {
```
craigatgoogle: Same comment as above, but additionally verifying correct propagation of the user defined service ID.

pkg/broker/registry.go
```diff
@@ -77,7 +86,9 @@ func MapServiceIdToName() map[string]string {
...
+		if entry, err := svc.CatalogEntry(); err == nil {
```
craigatgoogle: Not to dogpile onto this code, but you should probably handle the error case as well :)

josephlewis42: Fixed, I refactored to remove the need for this method. It was going to be tossed eventually so better sooner rather than later.

pkg/broker/broker_test.go
josephlewis42: Done, I really miss real set arithmetic. One of the few places I really think the go authors made a mistake.

PR #235, opened 2018-08-08
Custom Plan Tile Generation

josephlewis42:
Add the ability to generate service plan customization forms from the CLI.

Note1: I manually re-arranged the output of the service plan forms to move BigTable down next to Spanner so the diff would be easier to read. Normally they're in alphabetical order.

Note2: I decided the best way to get labels for these forms was to use the JSON name rather than adding a field for label into every variable which would introduce confusion.


brokerapi/brokers/cloudsql/postgres-definition.go
```diff
@@ -131,6 +131,30 @@ func init() {
 		ProvisionInputVariables: commonProvisionVariables,
 		BindInputVariables:      commonBindVariables,
 		BindOutputVariables:     commonBindOutputVariables,
+		PlanVariables: []broker.BrokerVariable{
+			broker.BrokerVariable{
+				FieldName: "tier",
+				Type:      broker.JsonTypeString,
+				Details:   "a string of the form db-custom-[CPUS]-[MEMORY_MBS], where memory is at least 3840",
```
craigatgoogle: Details description needs capitalization of the first word.

pkg/broker/registry.go
```diff
@@ -125,6 +127,17 @@ func (svc *BrokerService) UserDefinedPlansProperty() string {
 	return fmt.Sprintf("service.%s.plans", svc.Name)
 }
 
+// TileUserDefinedPlansVariable returns the name of the user defined plans
+// variable for the broker tile.
+func (svc *BrokerService) TileUserDefinedPlansVariable() string {
+	v := utils.PropertyToEnvUnprefixed(svc.Name)
+	if strings.HasPrefix(v, "GOOGLE_") {
+		v = v[7:]
```
craigatgoogle: This seems brittle. Consider instead:
prefix := "GOOGLE_"
if strings.HasPrefix(v, prefix) {
  v = v[len(prefix):]
}

pkg/broker/variables.go
```diff
@@ -34,4 +34,7 @@ type BrokerVariable struct {
 	Details string
 	// The default value of the field.
 	Default interface{}
+	// If there are a limited number of valid values for this field then
+	// Enum will hold them here along with friendly namers
```
craigatgoogle: typo? "namers"

PR #239, opened 2018-08-10
Customization Markdown Generation

josephlewis42:
Added generator code to generate the customization.md file. It has a lot of HTML to make the options easier to understand and to differentiate between the settings that are plain environment variables and those that need to be specified for custom plans.

The GitHub markdown parser doesn't play very nicely with markdown mixed with HTML so some of the stuff that looks like it could be done in markdown e.g. bolding "Required" doesn't render properly.

I also fixed a difference between manual configuration and tile configuration where you couldn't specify the database name if you were running the tile preventing multi-tenancy.


pkg/generator/customization-md.go
```diff
@@ -0,0 +1,168 @@
...
+If you are using the PCF tile, it will generate the UUIDs for you.
+DO NOT delete plans, instead you should change their labels to mark them as deprecated.
+
+{{ range $i, $f := .ServicePlanForms -}}
+	{{ template "customplanform" $f }}
+{{- end }}
+
+---------------------------------------
+
+_Note: **Do not edit this file**, it was auto-generated by running <code>gcp-service-broker generate customization</code>. If you find an error, change the source code in <tt>customization-md.go</tt> or file a bug._
```
evandbrown: Love the generation!

Issue #216, opened 2018-07-17
Abstract Database Layer


josephlewis42:
There are currently 46 locations where `DbConnection` is accessed outside of the `db_service` module. We should not be exposing direct connections to the database, instead DB access should be done through an abstraction layer to not rely on the underlying structure (or type) of the database.

PR #263, opened 2018-08-21
Generate Database DAOs

josephlewis42:
Create a `go generate`d DAO for our database models. This will allow us to remove most of the SQL statements in the codebase #216 in favor of access through a DAO which can eventually be mocked (or run on top of SQLite or stuck on Firestore/Datastore).

This is PR 1 of 2, laying the foundation with the code generation so we can next remove all the direct database access from the code.


josephlewis42:
The tests for this read more like integration tests per-DAO rather than unit tests for each method because it's so hard to test database operations in a vacuum. I can re-format if necessary but it'll add some verbosity.

db_service/dao_generator.go
```diff
@@ -0,0 +1,342 @@
...
+
+	// Should be able to create the item
+	beforeCreation := time.Now()
+	if err := ds.{{funcName "Create" .}}(&instance); err != nil {
+		t.Errorf("Expected to be able to create the item %#v, got error: %s", instance, err)
+	}
+	afterCreation := time.Now()
+
+	// after creation we should be able to get the item
+	ret, err := ds.{{funcName "Get" . .PrimaryKeyField}}(testPk)
```
craigatgoogle: Consider adding a test step which compares the equality of the retrieved object verses the original.

craigatgoogle:
Behavior driven testing is a feature in this case, so no worries about the unit tests. Just to re-state the TLDR from our 1:1, while I'm always a bit cautious to embrace code-generation unless it's absolutely necessary, I do see your point about Golang's lack of generics making the alternative very repetitive. I think adding the generated tests is a nice middle ground.

db_service/dao_generator.go
josephlewis42: It looks like the timezones are changed between serialization/deserialization even if the times (I _hate_ dates) are the same so I can't `.DeepEqual`, what do you think about me adding a `map[string]interface{}` to the generator for field/example pairs and do a field-by-field set and comparison from those? It feels kinda wonky because now the examples won't be type checked up until `go generate` is run but it feels better than reflection in templates.

db_service/dao_test.go
```diff
@@ -0,0 +1,352 @@
...
+// Code generated by go generate; DO NOT EDIT.
+// This file was generated at 2018-08-21 13:20:42.218760018 -0700 PDT m=+0.010055154.
+
+package db_service
+
+import (
+	"testing"
+	"time"
+
+  "github.com/GoogleCloudPlatform/gcp-service-broker/brokerapi/brokers/models"
```
evandbrown: Formatting problem

db_service/dao_generator.go
evandbrown: Is gorm messing with `time.Time` fields on deserialization?

josephlewis42: The times do represent the same point in time, but aren't created/deserialized in the same way so they can't be tested using reflection.

If I inject this line here's the output:

```
// t.Errorf("test %#v %#v, %#v", instance.CreatedAt.Equal(ret.CreatedAt), instance.CreatedAt, ret.CreatedAt)

dao_test.go:79: test true time.Time{wall:0xbed77c4956a625a3, ext:21705095, loc:(*time.Location)(0xfb4740)}, time.Time{wall:0x16a625a3, ext:63670546597, loc:(*time.Location)(0xc42046a600)}
```

The strings are something like the following: `2018-08-22 07:54:10.142900966 -0700 PDT m=+0.009016746` vs `2018-08-22 07:54:10.142900966 -0700 -0700`.

So I think the best courses of action are to either defer testing the serialization/deserialization to an integration test when it's used by another module or specify values field-by-field and test those explicitly which would keep it part of the code generation so it wouldn't be forgotten in future additions but still be flexible-ish enough.

craigatgoogle: A field by field test should work. One option to potentially make the code less verbose and more general is to have the models maintain a list of fields to exclude from the equality test. Here's an example of using reflect which could potentially be useful in this scenario: https://gist.github.com/drewolson/4771479
YMMV :)

PR #270, opened 2018-08-29
Fix and add godoc strings for the brokers

josephlewis42:
This doesn't fix all the missing strings across the application, but at least covers the brokers


josephlewis42:
@mattysweeps, updated. I also changed some of the function signatures/constants to private where they weren't used outside of the same package.

PR #275, opened 2018-09-05
Custom key DAO

### LINKED ISSUE #216: Abstract Database Layer

josephlewis42:
The DAO generator can now create functions for multi-column keys and generate tests for them. The multi-key approach really stretched the capabilities of the templates, but this should be the complete extent of what we need.

This commit also moves every remaining access through the database object to the DAO (except for the code in the database package, obviously).

Fixes #216


db_service/dao.go
```diff
@@ -13,19 +13,19 @@
 // limitations under the License.
```
josephlewis42: This file is auto-generated. You'll see some differences in pk -> id because ids are now treated the same as the multi-column keys.

db_service/dao.go
```diff
@@ -13,19 +13,19 @@
...
 
 // Code generated by go generate; DO NOT EDIT.
-// This file was generated at 2018-08-22 08:41:25.282061797 -0700 PDT m=+0.027525077.
```
josephlewis42: Removed these lines so running `go generate` will be idempotent which is useful because we have other generated code in the project too.

db_service/dao_test.go
```diff
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 // Code generated by go generate; DO NOT EDIT.
-// This file was generated at 2018-08-22 08:41:25.285530014 -0700 PDT m=+0.030993229.
```
josephlewis42: auto-generated file.

integration_tests/integration_test.go
```diff
@@ -94,8 +91,8 @@ func testGenericService(brokerConfig *config.BrokerConfig, gcpBroker *GCPService
 	Expect(err).ToNot(HaveOccurred())
 
 	// Provision is registered in the database
-	var count int
-	db_service.DbConnection.Model(&models.ServiceInstanceDetails{}).Where("id = ?", params.instanceId).Count(&count)
+	count, err := db_service.CountServiceInstanceDetailsById(params.instanceId)
```
josephlewis42: All of these CountXByYAndZ and CheckDeletedXByYAndZ could probably be rolled into a single "Exists" checker which would be a special case of "GetXByYAndZ", but that's for a later time.

brokerapi/brokers/cloudsql/broker.go
```diff
@@ -314,10 +314,11 @@ func (b *CloudSQLBroker) finishProvisioning(instanceId string, params map[string
 	// it should be quick and not actually async.
 	var err error
 
-	instance := models.ServiceInstanceDetails{}
-	if err = db_service.DbConnection.Where("ID = ?", instanceId).First(&instance).Error; err != nil {
+	instancePtr, err := db_service.GetServiceInstanceDetailsById(instanceId)
+	if err != nil {
 		return brokerapi.ErrInstanceDoesNotExist
 	}
+	instance := *instancePtr
```
mattysweeps: I see a couple of places where you duplicate the struct. I'm assuming this is to ensure the same functionality as before.

However, I think it's not necessary. If you follow this code, you can see that even if called sub-functions edit the struct, the final struct passed to the save call is still the copy. So you should be safe to just pass the pointer around.

brokerapi/brokers/cloudsql/broker.go
```diff
@@ -427,10 +428,11 @@ func (b *CloudSQLBroker) ensureUsernamePassword(instanceID, bindingID string, de
 // The default PCF service broker timeout may need to be raised to 90 or 120 seconds to accommodate the long bind time.
 func (b *CloudSQLBroker) Bind(instanceID, bindingID string, details brokerapi.BindDetails) (models.ServiceBindingCredentials, error) {
 
-	cloudDb := models.ServiceInstanceDetails{}
-	if err := db_service.DbConnection.Where("ID = ?", instanceID).First(&cloudDb).Error; err != nil {
+	cloudDbPtr, err := db_service.GetServiceInstanceDetailsById(instanceID)
+	if err != nil {
 		return models.ServiceBindingCredentials{}, brokerapi.ErrInstanceDoesNotExist
 	}
+	cloudDb := *cloudDbPtr
```
mattysweeps: Same comment

brokerapi/brokers/cloudsql/broker.go
```diff
@@ -533,7 +535,7 @@ func (b *CloudSQLBroker) PollInstance(instanceId string) (bool, error) {
 		return false, err
 	}
 
-	return b.PollOperation(instance, op)
+	return b.PollOperation(*instance, op)
```
mattysweeps: It's unfortunate `PollOperation` doesn't take a pointer. Looking at the function, it seems read only. Would it be possible to update it?

mattysweeps: I'm seeing a lot of instances of struct passing of `models.ServiceInstanceDetails` 

Maybe it would be better if we created an issue and handled this in another PR?

craigatgoogle:
LGTM, modulo mattysweeps feedback.

brokerapi/brokers/cloudsql/broker.go
```diff
@@ -524,23 +522,17 @@ func (b *CloudSQLBroker) Unbind(creds models.ServiceBindingCredentials) error {
 
 // PollInstance gets the last operation for this instance and checks its status.
 func (b *CloudSQLBroker) PollInstance(instanceId string) (bool, error) {
-
-	instance, err := db_service.GetServiceInstanceDetailsById(instanceId)
-	if err != nil {
-		return false, brokerapi.ErrInstanceDoesNotExist
-	}
-
```
josephlewis42: This is already checked by the code that calls this method, so no need to validate again.

josephlewis42:
@mattysweeps, updated PTAL when you get a chance.

Thanks!

Issue #251, opened 2018-08-14
Support Resource Labels


josephlewis42:
We should support setting labels for supported services.

This will help us tear down test environments more easily, and could provide useful information to operators if we included the org, space and provision/bind id on every resource.

josephlewis42:
All supported services have auto-generated PCF labels now if the org/space come in on the request (Bigtable is missing, but that's because the client library _still_ doesn't have support for them). Service account only projects also are missing labels.

#302 will handle user-defined labels.

PR #276, opened 2018-09-10
added label support for bigquery, cloudsql, spanner and gcs

josephlewis42:
Related to #251. Add labels to services that support them indicating the organization ID, space ID and instance ID to better support billing.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -18,9 +18,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"regexp"
 	"strings"
 
 	"github.com/GoogleCloudPlatform/gcp-service-broker/brokerapi/brokers/models"
+	"github.com/pivotal-cf/brokerapi"
 	"github.com/spf13/viper"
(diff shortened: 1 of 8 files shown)


utils/utils.go
```diff
@@ -173,3 +179,33 @@ func GetDefaultProjectId() (string, error) {
...
+		labels["pcf-organization-guid"] = orgGuid
+	}
+
+	if spaceGuid, ok := requestContext["space_guid"]; ok {
+		labels["pcf-space-guid"] = spaceGuid
+	}
+
+	sanitized := map[string]string{}
+	for key, value := range labels {
+		sanitized[key] = invalidLabelChars.ReplaceAllString(value, "_")
```
mattysweeps: Would it be a good idea to log when a key needed to be sanitized?

josephlewis42:
@mattysweeps added documentation about billing details. PTAL when you get a chance!

utils/utils.go
josephlewis42: I added some documentation about billing to include this now.

PR #282, opened 2018-09-17
don't use panics as flow-control

### LINKED ISSUE #228: Don't `panic()`
Several of our current execution error paths end in `panic` where they should end in `log.Fatalf()` instead because they're not recoverable errors and stack traces are unfriendly toward users. `grep -r "panic" . | grep -v "./vendor"` shows:

```
./db_service/db_service.go:			panic(fmt.Sprintf("Error migrating database: %s", err.Error()))
./db_service/setup_db.go:		panic("DB_HOST, DB_USERNAME and DB_PASSWORD are required environment variables.")
./brokerapi/brokers/models/db.go:		panic(err)
./brokerapi/brokers/models/db.go:			panic(err)
./brokerapi/brokers/broker_base/broker_base.go:	panic("Can't check last operation on a synchronous service")
```

The last one SHOULD NOT cause a panic, but instead an error message. The connection should error rather than crash due to a user mistake.

josephlewis42:
Fixes #228. Rather than using `panic` as a flow-control device, use `error` instead.

This mostly affects tests, which I've updated and re-ran to validate.

### DIFF
--- a/brokerapi/brokers/models/db.go
+++ b/brokerapi/brokers/models/db.go
@@ -23,7 +23,7 @@ import (
 type ServiceBindingCredentials ServiceBindingCredentialsV1
 
 // GetOtherDetails returns an unmarshaled version of the OtherDetails field
-// or panics.
+// or errors.
 func (sbc ServiceBindingCredentials) GetOtherDetails() (map[string]string, error) {
 	var creds map[string]string
 	err := json.Unmarshal([]byte(sbc.OtherDetails), &creds)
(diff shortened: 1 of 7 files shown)


brokerapi/brokers/models/db.go
```diff
@@ -24,30 +24,26 @@ type ServiceBindingCredentials ServiceBindingCredentialsV1
 
 // GetOtherDetails returns an unmarshaled version of the OtherDetails field
 // or panics.
```
mattysweeps: or errors, (also I don't know if you need to say "or errors")

Issue #250, opened 2018-08-14
Refactor GCPServiceBroker API Handler


josephlewis42:
Right now there isn't a clear separation of duties between the HTTP handler and the service broker manager.

We should break it into the following two roles:

The service broker API handler:

* Resolves any database references
* Gathers project ID, service account, logging context, JWT, request context, plan info
* Validates IDs, variable types and variable existence
* Saves results to the database

The individual broker:

* Gets an object containing the project id, service account, logger, JWT, request context, plan info, and variable resolver
* Does any complex validation e.g. length checking, bounds checking, invalid combinations etc.
* Makes the request
* Returns the result

The service broker API will no longer need to instantiate every broker with a `broker_base.BrokerBase` type, instead the info will be passed along with every request opening us up to simpler validation and more complex interactions in the future like multiple projects or limited scope service accounts.

PR #284, opened 2018-09-18
pushed down context usage

josephlewis42:
Working towards #250, pass context along with our requests.

### DIFF
--- a/brokerapi/brokers/broker_base/broker_base.go
+++ b/brokerapi/brokers/broker_base/broker_base.go
@@ -37,18 +37,18 @@ type BrokerBase struct {
 // Bind creates a service account with access to the provisioned resource with
 // the given instance.
 func (b *BrokerBase) Bind(ctx context.Context, instanceID, bindingID string, details brokerapi.BindDetails) (models.ServiceBindingCredentials, error) {
-	return b.AccountManager.CreateCredentials(instanceID, bindingID, details, models.ServiceInstanceDetails{})
+	return b.AccountManager.CreateCredentials(ctx, instanceID, bindingID, details, models.ServiceInstanceDetails{})
 }
 
 // BuildInstanceCredentials combines the bind credentials with the connection
 // information in the instance details to get a full set of connection details.
-func (b *BrokerBase) BuildInstanceCredentials(bindDetails models.ServiceBindingCredentials, instanceDetails models.ServiceInstanceDetails) (map[string]string, error) {
-	return b.AccountManager.BuildInstanceCredentials(bindDetails, instanceDetails)
+func (b *BrokerBase) BuildInstanceCredentials(ctx context.Context, bindDetails models.ServiceBindingCredentials, instanceDetails models.ServiceInstanceDetails) (map[string]string, error) {
+	return b.AccountManager.BuildInstanceCredentials(ctx, bindDetails, instanceDetails)
 }
 
 // Unbind deletes the created service account from the GCP Project.
 func (b *BrokerBase) Unbind(ctx context.Context, creds models.ServiceBindingCredentials) error {
-	return b.AccountManager.DeleteCredentials(creds)
+	return b.AccountManager.DeleteCredentials(ctx, creds)
 }
 
 // PollInstance does nothing but return an error because Base services are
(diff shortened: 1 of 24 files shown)


brokerapi/brokers/broker_base/broker_base.go
```diff
@@ -35,7 +36,7 @@ type BrokerBase struct {
 
 // Bind creates a service account with access to the provisioned resource with
 // the given instance.
-func (b *BrokerBase) Bind(instanceID, bindingID string, details brokerapi.BindDetails) (models.ServiceBindingCredentials, error) {
+func (b *BrokerBase) Bind(ctx context.Context, instanceID, bindingID string, details brokerapi.BindDetails) (models.ServiceBindingCredentials, error) {
 	return b.AccountManager.CreateCredentials(instanceID, bindingID, details, models.ServiceInstanceDetails{})
```
craigatgoogle: The ctx should probably be passed down through the AccountManager.CreateCredentials() call as well. It looks like down the call stack in createServiceAccount() there's still a use of background ctx on line 171.

brokerapi/brokers/broker_base/broker_base.go
```diff
@@ -46,13 +47,13 @@ func (b *BrokerBase) BuildInstanceCredentials(bindDetails models.ServiceBindingC
 }
 
 // Unbind deletes the created service account from the GCP Project.
-func (b *BrokerBase) Unbind(creds models.ServiceBindingCredentials) error {
+func (b *BrokerBase) Unbind(ctx context.Context, creds models.ServiceBindingCredentials) error {
 	return b.AccountManager.DeleteCredentials(creds)
```
craigatgoogle: Should probably also pass the ctx down through DeleteCredentials as well.

brokerapi/brokers/cloudsql/broker.go
```diff
@@ -507,7 +506,7 @@ func (b *CloudSQLBroker) BuildInstanceCredentials(bindDetails models.ServiceBind
 }
 
 // Unbind deletes the database user, service account and invalidates the ssl certs associated with this binding.
-func (b *CloudSQLBroker) Unbind(creds models.ServiceBindingCredentials) error {
+func (b *CloudSQLBroker) Unbind(ctx context.Context, creds models.ServiceBindingCredentials) error {
 
 	err := b.AccountManager.DeleteCredentials(creds)
```
craigatgoogle: Same as above, re: pushing ctx down through AM calls.

brokerapi/brokers/cloudsql/broker.go
```diff
@@ -429,7 +428,7 @@ func (b *CloudSQLBroker) ensureUsernamePassword(instanceID, bindingID string, de
 // Bind creates a new username, password, and set of ssl certs for the given instance.
 // The function may be slow to return because CloudSQL operations are asynchronous.
 // The default PCF service broker timeout may need to be raised to 90 or 120 seconds to accommodate the long bind time.
-func (b *CloudSQLBroker) Bind(instanceID, bindingID string, details brokerapi.BindDetails) (models.ServiceBindingCredentials, error) {
+func (b *CloudSQLBroker) Bind(ctx context.Context, instanceID, bindingID string, details brokerapi.BindDetails) (models.ServiceBindingCredentials, error) {
 
 	cloudDb, err := db_service.GetServiceInstanceDetailsById(instanceID)
```
craigatgoogle: I think for consistency's sake it would also make sense to pass the ctx through to the db_service function calls. The fact that it uses GORM is an implementation detail which the rest of the system shouldn't be aware of. My experience with Go code bases is if you're going to use ctx, then it's best practice to pass it to all downstream calls.

Separately, it looks like GORM does not currently supports contexts, although they have an open feature request for it: https://github.com/jinzhu/gorm/issues/1231 which would make a case for pre-positioning this code base to take advantage of it when it does get implemented.

josephlewis42: Good idea, done!

brokerapi/brokers/cloudsql/broker.go
```diff
@@ -690,15 +687,15 @@ func updateOperationId(instance models.ServiceInstanceDetails, operationId strin
...
 		return fmt.Errorf(`Error saving instance details to database: %s. WARNING: this instance cannot be deprovisioned through cf.
 		Please contact your operator for cleanup`, err)
 	}
 	return nil
 }
 
 // LastOperationWasDelete checks if the last async operation was a deletion (as opposed to a provision).
-func (b *CloudSQLBroker) LastOperationWasDelete(instanceId string) (bool, error) {
+func (b *CloudSQLBroker) LastOperationWasDelete(ctx context.Context, instanceId string) (bool, error) {
 	op, err := db_service.GetLastOperation(instanceId)
```
craigatgoogle: Should probably pass it through to GetLastOperation as well.

josephlewis42: Thanks, I forgot we had one call that wasn't auto-generated :-)

Issue #254, opened 2018-08-14
Migrate Existing Brokers to Terraform


josephlewis42:
Existing brokers should be migrated to Terraform. For each existing broker, we'll create Terraform files for binding and provisioning compatible with the existing arguments.

The provision file will just need to be compatible with the existing resource provisioning variables.

The bind file will import the parent resource (where needed e.g. GCS bucket) and create the associated service account.

De-provisioning and unbinding will continue to operate through manual calls to the REST endpoints.

This isn't a 100% ideal transition, but will allow us to add new parameters we desperately need for #238 #23 #16 #9 #251 #209 #4 #170 #205 #43 #61 and #53 without breaking functionality for services users have already provisioned.

- [x] BigQuery
- [x] Bigtable
- [x] CloudSQL MySQL
- [ ] CloudSQL PostgreSQL
- [x] Cloud Storage
- [x] Dataflow (preview)
- [x] Dataproc (preview)
- [x] Datastore
- [x] Dialogflow (preview)
- [x] Firestore (preview)
- [x] Memorystore for Redis (preview)
- [x] ML APIs
- [x] PubSub
- [x] Spanner
- [x] Stackdriver Debugger
- [x] Stackdriver Monitoring (preview)
- [x] Stackdriver Trace
- [x] Stackdriver Profiler

josephlewis42:
This is being tracked in the other repository. Some services can't be migrated currently due to limitations in the GCP Terraform provider.

Issue #243, opened 2018-08-10
Standardize Default and Operator Overrides


josephlewis42:
As part of our path to supporting user-defined services we need the code to consistently handle mashing together variables from the plan, defaults, those the user specifies and (new feature) operator chosen defaults (overrides to the defaults, like choosing EU data-centers by default).

The layers will be applied as follows:

	Plan Variables > User Defined Variables > Operator Defined Defaults > Hard-Coded Defaults

We should be able to use a mix of viper and our existing form generation tech to create new forms for overriding specific variables to handle the latter two. We can potentially also use Viper to do the overlays/variable lookup and casting.

As part of this refactoring, we'll get proper typing of variables for free going forward (rather than everything being a string).

This will fix #98

Issue #224, opened 2018-07-31
Extract Length Constants


josephlewis42:
`brokerapi/brokers/account_managers/sql_account_manager.go` and a few other places in the code do checks to see if a variable is over length x and if so truncate it. These numbers should be extracted out into constants and checked to see if the API has been updated to work with longer values.

josephlewis42:
This work was done as part of the migration to JSONSchema for every service

Issue #79, opened 2016-12-12
Add validation for tile configuration forms


ccemeraldeyes:
As well as service account permissions and api enablement

josephlewis42:
Pending https://github.com/pivotal-cf/docs-tiledev/issues/33

josephlewis42:
From that response, not much more can be done in the tile. The software itself has been slowly modified to do validation on initialization.

PR #292, opened 2018-10-04
Schema support

josephlewis42:
Start adding JSON Schema support to the broker. This commit tests the waters for the general approach.

After annotating the rest of the fields, we'll be able to push validation up into the broker to fail faster, have more complete documentation, and (possibly) expose the schemas like the OSB Spec recommends.

This work also opens up the way for improving/implementing:

* #243 - we'll be able to validate operator overrides
* #254 - existing brokers will behave the same way as the new ones
* #251 - we can now support more interesting types
* #250 - we can start mashing up variables earlier because tests are explicit
* #79 - we can generate these based on the schema for some forms
* #224 - these constraints will be explicit in the schema and not tested/accommodated for in multiple places

### DIFF
--- a/brokerapi/brokers/bigquery/definition.go
+++ b/brokerapi/brokers/bigquery/definition.go
@@ -59,8 +59,21 @@ func init() {
 			{
 				FieldName: "name",
 				Type:      broker.JsonTypeString,
-				Details:   "The name of the BigQuery dataset. Must be alphanumeric (plus underscores) and must be at most 1024 characters long.",
+				Details:   "The name of the BigQuery dataset.",
 				Default:   "a generated value",
+				Constraints: map[string]interface{}{
+					"pattern":   "^[A-Za-z0-9_]+$",
+					"maxLength": 1024,
+				},
+			},
+			{
+				FieldName: "location",
+				Type:      broker.JsonTypeString,
+				Details:   "The location of the BigQuery instance.",
+				Default:   "US",
+				Constraints: map[string]interface{}{
+					"examples": []string{"US", "EU", "asia-northeast1"},
+				},
 			},
 		},
 		DefaultRoleWhitelist: roleWhitelist,
--- a/brokerapi/brokers/bigquery/definition.go
+++ b/brokerapi/brokers/bigquery/definition.go
@@ -73,6 +73,7 @@ func init() {
 				Details:   "The location of the BigQuery instance.",
 				Default:   "US",
 				Constraints: validation.NewConstraintBuilder().
+					Pattern("^[A-Za-z][-a-z0-9A-Z]+$").
 					Examples("US", "EU", "asia-northeast1").
 					Build(),
 			},
--- a/brokerapi/brokers/bigquery/definition.go
+++ b/brokerapi/brokers/bigquery/definition.go
@@ -17,6 +17,7 @@ package bigquery
 import (
 	accountmanagers "github.com/GoogleCloudPlatform/gcp-service-broker/brokerapi/brokers/account_managers"
 	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/broker"
+	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/validation"
 )
 
 func init() {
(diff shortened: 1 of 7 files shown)


brokerapi/brokers/bigquery/definition.go
```diff
@@ -59,8 +60,21 @@ func init() {
...
+					Pattern("^[A-Za-z0-9_]+$").
+					MaxLength(1024).
+					Build(),
+			},
+			{
+				FieldName: "location",
+				Type:      broker.JsonTypeString,
+				Details:   "The location of the BigQuery instance.",
+				Default:   "US",
+				Constraints: validation.NewConstraintBuilder().
```
craigatgoogle: If you're going to employ data validation above, why not do so consistently across fields?

pkg/broker/variables.go
```diff
@@ -37,4 +44,45 @@ type BrokerVariable struct {
 	// If there are a limited number of valid values for this field then
 	// Enum will hold them in value:friendly name pairs
 	Enum map[interface{}]string
+	// Constraints holds JSON Schema validations defined for this variable.
+	// Keys are valid JSON Schema validation keywords, and values are their
+	// associated values.
+	Constraints map[string]interface{}
+}
+
+// ToSchema converts the BrokerVariable into the value part of a JSON Scheama
```
craigatgoogle: s/Sheama/Schema

brokerapi/brokers/bigquery/definition.go
josephlewis42: Were you thinking about adding a pattern for the region?

I'm hesitant to add enums for the regions, because it would mean we'd have to update for any new that go online or offline, but I think patterns could work.

pkg/generator/types.go
```diff
@@ -185,9 +186,65 @@ func varNotes(variable broker.BrokerVariable) string {
...
+		// Validation keywords for objects
+		{validation.KeyMaxProperties, "The object must have at most %v properties."},
+		{validation.KeyMinProperties, "The object must have at least %v properties."},
+		{validation.KeyRequired, "The following properties are required: %v."},
+		{validation.KeyPropertyNames, "Property names must match the JSON Schema: `%+v`."},
+	}
+
+	var bullets []string
+	for _, formatter := range constraintFormatters {
+		if v, ok := schema[formatter.SchemaKey]; ok {
```
mattysweeps: Should we error or log on `!ok`? Just in case someone adds a constraint but forgets to add a new bullet?

pkg/generator/types.go
```diff
@@ -185,9 +186,65 @@ func varNotes(variable broker.BrokerVariable) string {
...
 	return out
 }
 
+// constraintsToDoc converts a map of JSON Schema validation key/values to human-readable bullet points.
+func constraintsToDoc(schema map[string]interface{}) []string {
+	// We use an anonymous struct rather than a map to get a strict ordering of
+	// constraints so they are generated consistently in documentation.
+	// Not all JSON Schema constraints can be cleanly expressed in this format,
+	// nor do we use them all so some are missing.
+	constraintFormatters := []struct {
```
craigatgoogle: This list of formatters seems like it only needs to be defined once. Consider defining as a const outside of the function, perhaps defined before the function ease of reference.

pkg/broker/variables.go
```diff
@@ -37,4 +44,45 @@ type BrokerVariable struct {
 	// If there are a limited number of valid values for this field then
 	// Enum will hold them in value:friendly name pairs
 	Enum map[interface{}]string
+	// Constraints holds JSON Schema validations defined for this variable.
+	// Keys are valid JSON Schema validation keywords, and values are their
```
craigatgoogle: Consider adding a link to the JSON Schema specification being employed here.

pkg/validation/constraint_builder.go
```diff
@@ -0,0 +1,185 @@
...
+)
+
+//  NewConstraintBuilder creates a builder for JSON Schema compliant constraint
+// lists.
+func NewConstraintBuilder() ConstraintBuilder {
+	return ConstraintBuilder{}
+}
+
+// A builder for JSON Schema compliant constraint lists
+type ConstraintBuilder map[string]interface{}
```
craigatgoogle: Another example of Golang's dire need for generics. </rant>

brokerapi/brokers/bigquery/definition.go
craigatgoogle: Yep, a pattern is what I had in mind :)

craigatgoogle:
LGTM. Just some minor notes.

pkg/generator/types.go
josephlewis42: I pulled it into a var and a struct, but can't do a const because I guess the structs aren't a `constexpr`.

pkg/generator/types.go
josephlewis42: I think it's a risk we'll have to take. There are some fields we don't want to document in this format (like name, type, and details which are accounted for in the section above the bullets) and I doubt the spec will change much, but it might. 

I left a comment up in the struct for us in the future to remember why not everything gets a bullet point.

pkg/generator/types.go
craigatgoogle: Bummer, in that case keeping it encapsulated inside the function is better from a defensive coding standpoint.

josephlewis42: moved it back

Issue #98, opened 2017-01-26
Configure defaults for provisioning request parameters


claudio-benfatto:
Hello,

first of all thanks for developing this project!

Our use case is the following:

we have most of our infrastructure deployed in the `eu-west` region and we would like to set the default provisioning parameters for some of the resources (for example `google-storage` and `google-cloudsql`) accordingly and at deployment time. 
Mainly, so that our users could be spared from setting those parameters which each request ie. 

```
cf create-service google-storage standard test-bucket-eu -c '{"location": "eu"}'`
 ```

what would be really nice in my opinion is to add the capability of overriding those hard-coded default values passing in the manifest some optional json (or any other format) configuration, ie.:

```
GOOGLE_STORAGE_PROVISIONING_DEFAULTS: '{"location": "eu"}'
```

Let me know what you think about this proposal... or maybe if there is already a way to do it, that perhaps I overlooked!

Thanks,
Claudio

ccemeraldeyes:
Great idea! I'll get this prioritized. Sorry for the late response!

claudio-benfatto:
Great! Thanks @ccemeraldeyes !

mgoddard-pivotal:
This same approach could be used to help make binding to the service instance simpler.  For example, if we envision doing this routinely:

  `cf bind-service myapp foobar -c '{"role": "pubsub.admin"}'`

then having this type of configuration available would simplify the binding:

  `GOOGLE_STORAGE_BINDING_DEFAULTS: '{"role": "pubsub.admin"}'`

ccemeraldeyes:
proposal:

* new table ServiceDefaults (Id, ServiceId, Type, Key, Value) - where Type = "provision" or "bind"
* new env variables GOOGLE_<service name>_DEFAULTS for each service

`GOOGLE_STORAGE_DEFAULTS: '[
        {"type": "provision", "key": "location", "value": "eu"},
        {"type": "bind", "key": "role", "value": "storage.admin"}
]'`
* new tile pages for each service (probably combine with #51 ) with dropdown prefilled with provision or bind, dropdown prefilled with available settings, and textbox for value

@claudio-benfatto @mgoddard-pivotal if this seems reasonable I'll start this today

mgoddard-pivotal:
Yes, @ccemeraldeyes, that would work nicely, I think.

ccemeraldeyes:
[...]

PR #294, opened 2018-10-08
Default Interpolation

josephlewis42:
This PR lays the groundwork for resolving variables in a consistent, repeatable and extensible way for both our existing brokers and those in the future which will be defined using text files rather than code.

It introduces three new concepts and implements them in a single broker.

The new concepts are:

1. An execution engine. The execution engine is based on Hashicorp's HIL, which itself is similar to Pivotal's SPEL. Now, when a service wants to have a computed default variable, it can express it using HIL, which will be directly rendered to documentation rather than the more opaque "a generated value".
2. A context builder. Handling for variables used during provisioning and binding used to be defined in a handful of places (request JSON, default functions, service-definition and/or provision details) leading to complex logic and lack of options. Now there's a unified variable "context" and lookup happens in a strict order (operator defined service plan vars -> user vars -> default vars), this will allow us more flexibility in the future, like operator plans that override some user-variables or operator-defined defaults.
3. A variable resolver, capable of casting variables into the right types and accumulating errors.

Eventual results of this work:

* Allow user-defined templates.
* Significant reduction in complexity (5-10 line reduction per variable, 1-2 if statements).
* Remove the `name_generator` package.
* Greater transparency for users.
* Allow operator defaults #98, #243 
* Pull the common code into the api handler as part of #250


josephlewis42:
I have a slight modification to this code to make some of the other services' defaults better, I'll check it in tomorrow sometime. But 90% of it will be the same.

Essentially refactoring out the builder construction part into the service definition and add a flag to defaults so they can either override an existing value or not (enabling some of the constants stuff we'll need when moving to Terraform).

pkg/varcontext/interpolation/funcs.go
```diff
@@ -0,0 +1,103 @@
...
+
+	"github.com/hashicorp/hil/ast"
+	"github.com/spf13/cast"
+)
+
+var hilStandardLibrary = createStandardLibrary()
+
+// createStandardLibrary instantiates all the functions and associates them
+// to their names in a lookup table for our standard library.
+func createStandardLibrary() map[string]ast.Function {
```
mattysweeps: I feel we should create another doc file that describes more about the HIL interpolation and also lists these functions. (Since I'm assuming these functions are the only ones that can be evaluated)

Especially if customers are going to be using these down the road.

pkg/varcontext/errorcollector.go
```diff
@@ -0,0 +1,54 @@
...
+
+// ErrorCollector is a utility class that concatenates errors.
+// It is useful for builder-style objects that want to keep errors tracked
+// internally until a finalization call.
+type ErrorCollector struct {
+	errs []error
+}
+
+// HasErrors returns true if the collector has one or more errors.
+func (ec *ErrorCollector) HasErrors() bool {
```
mattysweeps: `HasErrors` is useful for some cases, but I think it's overused, since calling `Error()` and checking for `!nil` answers the same question.

pkg/varcontext/builder.go
```diff
@@ -0,0 +1,102 @@
...
+	builder.AddError(json.Unmarshal(data, &out))
+	builder.MergeMap(out)
+
+	return builder
+}
+
+// Build generates a finalized VarContext based on the state of the builder.
+// Exactly one of VarContext and error will be nil.
+func (builder *ContextBuilder) Build() (*VarContext, error) {
+	if builder.HasErrors() {
```
mattysweeps: The following accomplishes the same thing without `HasErrors`

```
if err := builder.Error(); err != nil; {
  return nil, err
}
```

pkg/varcontext/varcontext.go
```diff
@@ -0,0 +1,72 @@
...
+
+package varcontext
+
+import (
+	"fmt"
+
+	"github.com/spf13/cast"
+)
+
+type VarContext struct {
```
mattysweeps: Making `VarContext` an `ErrorCollector` has the weird side effect of exposing the `AddError` call. It's not a big deal, but it means that a varcontext caller can just add arbitrary errors to it. Instead, if `VarContext` had an internal field `ErrorCollector`, all it should have to do is implement `Error()` by calling into the `ErrorCollector`. That would stop the accidental exposure and should cover all needs (assuming HasError calls have also been replaced, see other comments)

brokerapi/brokers/bigquery/broker.go
```diff
@@ -39,44 +39,39 @@ type InstanceInformation struct {
 
 // Provision creates a new BigQuery dataset from the settings in the user-provided details and service plan.
 func (b *BigQueryBroker) Provision(ctx context.Context, instanceId string, details brokerapi.ProvisionDetails, plan models.ServicePlan) (models.ServiceInstanceDetails, error) {
-	var err error
-	var params map[string]string
+	variableContext, err := varcontext.Builder().
+		MergeDefaults(serviceDefinition().ProvisionInputVariables).
```
mattysweeps: Is this the very specif order you call out in the PR comments? `(operator defined service plan vars -> user vars -> default vars)`

Is this ordering documented in a `.md` file somewhere?

pkg/varcontext/varcontext.go
```diff
@@ -0,0 +1,72 @@
...
+
+	if err := validator(val); err != nil {
+		vc.AddError(fmt.Errorf("value for %q must be a %s", key, typeName))
+	}
+}
+
+// GetString gets a string from the context, storing an error if the key doesn't
+// exist or the variable couldn't be converted to a string.
+func (vc *VarContext) GetString(key string) string {
+	vc.validate(key, "string", func(val interface{}) error {
```
craigatgoogle: It seems like this approach causes the value to be parsed twice each time. Consider using a closure to store the result from the anonymous function's parsing, and returning that value from the outer function.

pkg/varcontext/errorcollector.go
```diff
@@ -0,0 +1,54 @@
...
+	if len(ec.errs) == 0 {
+		return nil
+	}
+
+	var out []string
+	for _, err := range ec.errs {
+		out = append(out, err.Error())
+	}
+
+	return errors.New(strings.Join(out, ", "))
```
craigatgoogle: For improved readability during debugging, consider joining on '\n'.

craigatgoogle: On second pass, I think serializing all the errors to a string might be hazardous. Go allows custom defined error types which could potentially contain more information than just a string. This limits the ability of upstream logic to benefit from such behavior. While such functionality might not be currently utilized, I suggest staying clear of decisions like this which limit your ability to utilize such language constructs completely.  Hashicorp has a package which might be of use to you in this situation: https://github.com/hashicorp/go-multierror

pkg/varcontext/interpolation/funcs.go
josephlewis42: Added a new doc with info about hil and the available functions. This can be filled in more once we add TF support.

pkg/varcontext/varcontext.go
josephlewis42: Fixed

pkg/varcontext/builder_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/broker"
+)
+
+func TestContextBuilder(t *testing.T) {
+	cases := map[string]struct {
+		Builder     *ContextBuilder
+		Expected    map[string]interface{}
+		ErrContains string
+	}{
+		"an empty context": {
```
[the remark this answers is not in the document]
josephlewis42: I refactored this a bit and added more tests here and a big one for the whole context eval/overloading in the context of a mock service in pkg/broker.

josephlewis42:
Updated to include docs on the functions and variable resolutions, tests around the interaction between the different variable sources, and a slightly modified way to compute default variables.

There's also now a "computed" variable section that gets executed at the end. This will be useful for things like merging user-defined labels with org/space labels, upgrading legacy services, and/or allowing constants to be injected into Terraform variables files (see the internal GCP Service Broker go-forward doc Appendix A for why we will need those).

pkg/varcontext/varcontext.go
```diff
@@ -70,3 +73,22 @@ func (vc *VarContext) ToMap() map[string]interface{} {
 
 	return output
 }
+
+// Error gets the accumulated error(s) that this VarContext holds.
+func (vc *VarContext) Error() error {
+	if vc.errors == nil {
```
mattysweeps: Same question about this being `errors` vs `errors.Error()`

pkg/varcontext/builder.go
```diff
@@ -94,8 +108,9 @@ func (builder *ContextBuilder) MergeJsonObject(data json.RawMessage) *ContextBui
 // Build generates a finalized VarContext based on the state of the builder.
 // Exactly one of VarContext and error will be nil.
 func (builder *ContextBuilder) Build() (*VarContext, error) {
-	if builder.HasErrors() {
-		return nil, builder.Error()
+	if builder.errors != nil {
```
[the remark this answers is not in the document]
josephlewis42: I think this one is correct, `builder.errors` is an error type that becomes non-nil the first time an error is "appended" to it, so `.Error()` would return a string.

I didn't like all the type casting multierror does, so I used the multierror.error type directly; which also makes the code more explicit.

pkg/varcontext/varcontext.go
```diff
@@ -0,0 +1,97 @@
...
+	if err := validator(val); err != nil {
+		vc.errors = multierror.Append(vc.errors, fmt.Errorf("value for %q must be a %s", key, typeName))
+	}
+}
+
+// GetString gets a string from the context, storing an error if the key doesn't
+// exist or the variable couldn't be converted to a string.
+func (vc *VarContext) GetString(key string) string {
+	var res string
+	var err error
```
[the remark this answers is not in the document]
josephlewis42: updated

docs/custom-services.md
```diff
@@ -0,0 +1,44 @@
...
+## Expression language reference
+
+The broker uses the [HIL expression language](https://github.com/hashicorp/hil) with a limited set of built-in functions.
+
+## Function reference
+
+The following string interpolation functions are available for use:
+
+* `time.nano() -> string`
+  * This function returns the current UNIX time in nanoseconds as a decimal string.
```
craigatgoogle: It could be worth mentioning that this is the UNIX time from the VM the SB is running on.

PR #297, opened 2018-10-09
migrated Bigtable to using templated vars

josephlewis42:
Migrated Bigtable to using templated vars and added tests and JSON Schema. I grabbed constraints from the Bigtable UI after trying to find an api.json in the google-cloud-go library, api.yaml in magic-modules, the docs, and the API docs based on the protobufs.


brokerapi/brokers/bigtable/broker.go
```diff
@@ -48,81 +46,50 @@ var storageTypes = map[string]googlebigtable.StorageType{
...
-	}
-	if userClusterId, clusterIdOk := params["cluster_id"]; clusterIdOk {
-		clusterId = userClusterId
-	}
-
-	numNodes, err := strconv.Atoi(plan.ServiceProperties["num_nodes"])
-	if err != nil {
-		return models.ServiceInstanceDetails{}, fmt.Errorf("Error converting num_nodes to int: %s", err)
-	}
+	instanceName := provisionContext.GetString("name")
```
craigatgoogle: Does instanceName need to be checked for an empty value before being used?

josephlewis42: It will be once #303 is done, and the API client will catch it, but I'll add a friendly error in the meantime.

PR #298, opened 2018-10-10
spanner interpolation work

josephlewis42:
This brings the interpolation constructs to Spanner.
Along with that and the regex, this refactor brings an additional change to make us more OSB compliant.

Rather than storing cloud operations in a separate table and relying on custom logic, this commit adds a generic field to hold an operation ID and instance state in the ServiceInstanceDetails table.

This has a few benefits:

* We can remove the CloudOperations table and DAO once CloudSQL is moved over to using the same technique.
* Multiple kinds of back-ends can use this state (like TF or DeploymentManager).
* Now, the service broker will return an OperationData field when a service was provisioned asynchronously bringing us into line with the spec.
* We can delete the `LastOperationWasDelete` call that had to be implemented (or hard-coded) for _every_ service.
* We can start to support update operations if we don't assume states as only provision or delete.
* We can use the presence of the field as a lock, knowing it's transactional with updates to the object whereas a separate table has a bunch of weird states the application can get into.


brokerapi/brokers/models/db.go
```diff
@@ -30,8 +41,8 @@ func (sbc ServiceBindingCredentials) GetOtherDetails() (map[string]string, error
 	return creds, err
 }
 
-// ServiceInstanceDetails holds information about provisioned services 
-type ServiceInstanceDetails ServiceInstanceDetailsV1
+// ServiceInstanceDetails holds information about provisioned services
```
craigatgoogle: Comment is missing a period.

brokerapi/brokers/spanner/broker.go
```diff
@@ -45,182 +43,104 @@ type InstanceInformation struct {
...
-	} else if spannerInstance == nil && err == nil && !done {
-		op.Status = string(instancepb.Instance_STATE_UNSPECIFIED)
 
-		if err := db_service.SaveCloudOperation(ctx, op); err != nil {
-			return false, fmt.Errorf(`Error saving operation details to database: %s.`, err)
+	case err == nil && done: // The operation was successful
+		instance.OperationId = ""
+		instance.OperationType = models.ClearOperationType
+		if err := db_service.SaveServiceInstanceDetails(ctx, instance); err != nil {
+			s.Logger.Error("updating instance after provision", err)
```
craigatgoogle: This block seems to swallow the error and not return it. That seems contradictory with the design of the rest of the method.

brokerapi/brokers/spanner/definition.go
```diff
@@ -63,20 +70,36 @@ func init() {
...
 			},
 			{
 				FieldName: "location",
 				Type:      broker.JsonTypeString,
 				Default:   "regional-us-central1",
-				Details:   `The location of the Spanner instance.`,
+				Details: `A configuration for a Cloud Spanner instance.
+				 Configurations define the geographic placement of nodes and their replication and are slightly different from zones.
+				 There are single region configurations, multi-region configurations, and multi-continent configurations.
+				 See the instance docs for a list of configurations: https://cloud.google.com/spanner/docs/instances.`,
```
craigatgoogle: Consider removing the period at the end of this sentence to prevent users from unwittingly copying/pasting it and getting a 404.

brokerapi/brokers/spanner/definition.go
```diff
@@ -63,20 +70,36 @@ func init() {
...
+				Constraints: validation.NewConstraintBuilder().
+					Examples("regional-asia-east1	", "nam3", "nam-eur-asia1").
```
craigatgoogle: Is the whitespace after regional-asia-east1 intentional?

db_service/migrations.go
```diff
@@ -29,7 +29,7 @@ import (
 // runs schema migrations on the provided service broker database to get it up to date
 func RunMigrations(db *gorm.DB) error {
 
-	migrations := make([]func() error, 3)
+	migrations := make([]func() error, 4)
```
craigatgoogle: Hard-coded magic numbers like these set off my CS TA alarm bells :)

pkg/varcontext/varcontext.go
```diff
@@ -58,6 +58,7 @@ func (vc *VarContext) GetInt(key string) (res int) {
 		return err
 	})
 
+
```
craigatgoogle: Double checking that this added whitespace was intentional.

josephlewis42: It was not, just a merge mistake.

PR #299, opened 2018-10-10
Pub/Sub interpolation

josephlewis42:
- Pub/Sub now uses interpolation. #250
- Pub/Sub now adds default labels to created topics and subscriptions. #251
- New validation documentation for Pub/Sub. #243
- New `GetBool` type on VariableContext.

### DIFF
--- a/brokerapi/brokers/pubsub/broker.go
+++ b/brokerapi/brokers/pubsub/broker.go
@@ -19,12 +19,11 @@ import (
 
 	googlepubsub "cloud.google.com/go/pubsub"
 	"github.com/GoogleCloudPlatform/gcp-service-broker/brokerapi/brokers/models"
-	"github.com/GoogleCloudPlatform/gcp-service-broker/brokerapi/brokers/name_generator"
+	"github.com/GoogleCloudPlatform/gcp-service-broker/utils"
 	"github.com/pivotal-cf/brokerapi"
 	"golang.org/x/net/context"
 
 	"fmt"
-	"strconv"
 	"time"
 
 	"github.com/GoogleCloudPlatform/gcp-service-broker/brokerapi/brokers/broker_base"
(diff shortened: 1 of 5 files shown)


brokerapi/brokers/pubsub/broker.go
```diff
@@ -47,79 +46,72 @@ type InstanceInformation struct {
...
-		if err != nil {
+		if _, err := pubsubClient.CreateSubscription(ctx, subscriptionName, subscriptionConfig); err != nil {
 			return models.ServiceInstanceDetails{}, fmt.Errorf("Error creating subscription: %s", err)
 		}
+	}
 
-		ii.SubscriptionName = params["subscription_name"]
+	ii := InstanceInformation{
+		TopicName:        topicName,
+		SubscriptionName: subscriptionName,
```
craigatgoogle: I noticed that subscriptionName is being used here without an empty string check, whereas previously in the function it's checked before use. I just wanted to sanity check and make sure this was intentional.

brokerapi/brokers/pubsub/broker.go
```diff
@@ -47,79 +46,72 @@ type InstanceInformation struct {
...
 
 	otherDetails, err := json.Marshal(ii)
 	if err != nil {
 		return models.ServiceInstanceDetails{}, fmt.Errorf("Error marshalling json: %s", err)
 	}
-	i.OtherDetails = string(otherDetails)
 
-	return i, nil
+	return models.ServiceInstanceDetails{
+		Name:         topicName,
```
craigatgoogle: It appears topicName is also not checked for an empty string before being used.

pkg/varcontext/varcontext.go
```diff
@@ -61,6 +61,19 @@ func (vc *VarContext) GetInt(key string) (res int) {
 	return
 }
 
+// GetBool gets a boolean from the context, storing an error if the key doesn't
+// exist or the variable couldn't be converted to a bool.
+// Integers can behave like bools in C style, 0 is false.
+// The strings "true" and "false" are also cast to their bool values.
+func (vc *VarContext) GetBool(key string) (res bool) {
+	vc.validate(key, "boolean", func(val interface{}) (err error) {
+		res, err = cast.ToBoolE(val)
```
craigatgoogle: Nice use of a closure and return var together :)

brokerapi/brokers/pubsub/broker.go
josephlewis42: It is intentional, but I'll add a comment up in the InstanceInformation definition saying it can be blank.

brokerapi/brokers/pubsub/broker.go
josephlewis42: This is semi-intetional. @mattysweeps is going to implement JSON Schema validation of user arguments some time this week, but I'll add the validation back in here so master is valid until then.

craigatgoogle: Sounds good. Thanks.

PR #307, opened 2018-10-16
move handling of async deprovision to GCPServiceBroker

josephlewis42:
This moves handling of the operation ID for asynchronous deprovision calls into GCPServiceBroker.

The service broker is responsible for managing the state of the instance, platform, and database, allowing the ServiceBrokerHelper to be responsible for just calling the right APIs and formatting requests.

part of #250 and #305

### DIFF
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -122,6 +122,8 @@ user-defined plans.
 			if err := client.RunExamplesForService(apiClient, serviceName); err != nil {
 				log.Fatalf("Error executing examples: %v", err)
 			}
+
+			log.Println("Success")
 		},
 	}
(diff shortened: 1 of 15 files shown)


cmd/client.go
```diff
@@ -122,6 +122,8 @@ user-defined plans.
 			if err := client.RunExamplesForService(apiClient, serviceName); err != nil {
 				log.Fatalf("Error executing examples: %v", err)
 			}
+
+			log.Println("Success")
```
josephlewis42: This is just a long-standing deficiency. No more needing to call `$?` to double-check the exit-status.

PR #308, opened 2018-10-16
update service state after successful lastoperation

josephlewis42:
Add state polling to the BrokerHelper explicitly to be called after LastOperation succeeds. This does three things:

* Removes an additional piece of lifecycle management logic from each broker.
* Provides a standardized way we can upgrade state in the future when we need additional info to be gathered/stored. For example, re-fetching CloudSQL's status after an upgrade from v1 to v2.
* Gives us a path towards supporting updates.

Part of #250 and #305

### DIFF
--- a/brokerapi/brokers/cloudsql/broker.go
+++ b/brokerapi/brokers/cloudsql/broker.go
@@ -453,12 +453,14 @@ func (b *CloudSQLBroker) PollInstance(ctx context.Context, instance models.Servi
 	}
 
 	if instance.OperationType == models.ProvisionOperationType {
-		instancePtr := &instance
-		if err := b.UpdateInstanceDetails(ctx, instancePtr); err != nil {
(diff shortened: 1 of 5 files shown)


brokerapi/brokers/cloudsql/broker.go
```diff
@@ -453,19 +453,20 @@ func (b *CloudSQLBroker) PollInstance(ctx context.Context, instance models.Servi
 	}
 
 	if instance.OperationType == models.ProvisionOperationType {
-		if err := b.refreshServiceInstanceDetails(ctx, &instance); err != nil {
+		instancePtr := &instance
```
craigatgoogle: Referencing a variable should be an O(1) op, so there's not performance benefit from storing the pointer. "&instance" on its own is pretty clear to read whereas using instancePtr requires someone to lookup the variable definition/instantiation to understand the code.

PR #309, opened 2018-10-17
Validate provision variables

mattysweeps:
Uses the gojsonschema library to validate provision parameters against
their respective BrokerVariable JSONSchema.

### DIFF
--- a/pkg/broker/variables.go
+++ b/pkg/broker/variables.go
@@ -19,6 +19,10 @@ import (
 	"sort"
 
 	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/validation"
+	"github.com/xeipuuv/gojsonschema"
+	"github.com/hashicorp/go-multierror"
+	"errors"
+	"strings"
 )
 
 const (
@@ -86,4 +90,63 @@ func (bv *BrokerVariable) ToSchema() map[string]interface{} {
 	}
 
 	return schema
+}
+
+// ValidateVariables validates a list of BrokerVariables adhere to their JSONSchema.
+func ValidateVariables(parameters map[string]interface{}, schemaVariables []BrokerVariable) error {
+
+	allErrors := &multierror.Error{
+		ErrorFormat:lineErrorFormatter,
+	}
+
+	for _, variable := range schemaVariables {
+
+		value, ok := parameters[variable.FieldName]
+		if !ok {
+			// According to json schema, the required property trumps the default value.
+			if variable.Required {
+				multierror.Append(allErrors, fmt.Errorf("missing required parameter \"%s\"", variable.FieldName))
+				continue
+			}
+
+			if variable.Default == nil {
+				continue
+			}
+
+			// Insert the default value into the parameters
+			value = variable.Default
+			parameters[variable.FieldName] = value
+		}
+
+		result, err := gojsonschema.Validate(gojsonschema.NewGoLoader(variable.ToSchema()), gojsonschema.NewGoLoader(value))
+		if err != nil {
+			multierror.Append(allErrors, err)
+			continue
+		}
+
+		if len(result.Errors()) > 0 {
+			for _, r := range result.Errors() {
+				// For better output, replace the "root" keyword for the root json object to the variable name.
+				multierror.Append(allErrors, errors.New(strings.Replace(r.String(), "(root)", fmt.Sprintf("(%s)", variable.FieldName), -1)))
+			}
+
+			continue
+		}
+
+	}
+
+	if len(allErrors.Errors) == 0 {
+		return nil
+	}
+
+	return allErrors
+}
+
+func lineErrorFormatter(es []error) string {
+	points := make([]string, len(es))
+	for i, err := range es {
+		points[i] = err.Error()
+	}
+
+	return fmt.Sprintf("%d error(s) occurred:\n%s", len(es), strings.Join(points, "\n"))
 }
--- a/pkg/broker/variables_test.go
+++ b/pkg/broker/variables_test.go
@@ -17,6 +17,8 @@ package broker
 import (
 	"reflect"
 	"testing"
+	"errors"
+	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/validation"
 )
 
 func TestBrokerVariable_ToSchema(t *testing.T) {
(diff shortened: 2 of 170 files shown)


pkg/broker/variables.go
```diff
@@ -87,3 +92,43 @@ func (bv *BrokerVariable) ToSchema() map[string]interface{} {
...
+			}
+		}
+
+		result, err := gojsonschema.Validate(gojsonschema.NewGoLoader(variable.ToSchema()), gojsonschema.NewGoLoader(value))
+		if err != nil {
+			multierror.Append(allErrors, err)
+		} else if len(result.Errors()) > 0 {
+			for _, r := range result.Errors() {
+				// For better output, replace the "root" keyword for the root json object to the variable name.
+				multierror.Append(allErrors, errors.New(strings.Replace(r.String(), "(root)", fmt.Sprintf("(%s)", variable.FieldName), -1)))
```
josephlewis42: I think it would make sense to construct a JSONSchema for the whole object and validate that. That way the schema will handle the required fields and I think you won't have to replace (root) because it should show the correct property name:

```
var required []string
properties := make(map[string]interface{})

for _, variable := range schemaVariables {
	properties[variable.FieldName] = variable.ToSchema()
	if variable.Required() {
		required = append(required, variable.FieldName)
	}
}

objectSchema := map[string]interface{}{
	"required": required,
	"properties": properties,
}

result, err := gojsonschema.Validate(gojsonschema.NewGoLoader(objectSchema), gojsonschema.NewGoLoader(parameters))
//...
```

pkg/broker/variables.go
```diff
@@ -87,3 +92,43 @@ func (bv *BrokerVariable) ToSchema() map[string]interface{} {
...
+
+		value, ok := parameters[variable.FieldName]
+		if !ok {
+			// According to json schema, the required property trumps the default value.
+			if variable.Required {
+				multierror.Append(allErrors, fmt.Errorf("missing required parameter \"%s\"", variable.FieldName))
+				continue
+			} else if variable.Default != nil {
+				// Insert the default value into the parameters
+				value = variable.Default
```
josephlewis42: I don't think the defaults will always end up being valid because some might be templates that are executed.

pkg/broker/variables_test.go
```diff
@@ -80,3 +82,164 @@ func TestBrokerVariable_ToSchema(t *testing.T) {
...
+					FieldName:"test",
+					Type:JsonTypeInteger,
+				},
+				{
+					Required:true,
+					FieldName:"test2",
+					Type:JsonTypeInteger,
+				},
+			},
+			Expected: errors.New("2 error(s) occurred: (test): Invalid type. Expected: integer, given: string; (test2): Invalid type. Expected: integer, given: string"),
```
josephlewis42: These are really nice error messages!

pkg/varcontext/builder.go
```diff
@@ -109,7 +110,7 @@ func (builder *ContextBuilder) MergeJsonObject(data json.RawMessage) *ContextBui
 // Exactly one of VarContext and error will be nil.
 func (builder *ContextBuilder) Build() (*VarContext, error) {
 	if builder.errors != nil {
-		builder.errors.ErrorFormat = lineErrorFormatter
+		builder.errors.ErrorFormat = utils.LineErrorFormatter
```
josephlewis42: Hooray for code reuse!

utils/utils.go
```diff
@@ -209,3 +209,12 @@ func ExtractDefaultLabels(instanceId string, details brokerapi.ProvisionDetails)
 
 	return sanitized
 }
+
+func LineErrorFormatter(es []error) string {
```
josephlewis42: This should probably have a godoc

pkg/broker/variables.go
mattysweeps: This is done, created a helper function `createJsonSchema`

pkg/broker/variables.go
mattysweeps: Currently defaults are applied to missing variables, then the whole map is ran through the json schema validation.

utils/utils.go
mattysweeps: Renamed this to `SingleLineErrorFormatter` and added godoc.

pkg/broker/variables.go
```diff
@@ -87,3 +91,61 @@ func (bv *BrokerVariable) ToSchema() map[string]interface{} {
...
+
+	for _, v := range variables {
+		if _, ok := parameters[v.FieldName]; !ok && v.Default != nil {
+			parameters[v.FieldName] = v.Default
+		}
+	}
+
+}
+
+// ValidateVariables validates a list of BrokerVariables adhere to their JSONSchema.
```
craigatgoogle: s/adhere/adhering

pkg/broker/variables.go
```diff
@@ -87,3 +91,61 @@ func (bv *BrokerVariable) ToSchema() map[string]interface{} {
...
+	allErrors := &multierror.Error{
+		ErrorFormat:utils.SingleLineErrorFormatter,
+	}
+
+	schema := createJsonSchema(variables)
+	result, err := gojsonschema.Validate(gojsonschema.NewGoLoader(schema), gojsonschema.NewGoLoader(parameters))
+
+	if err != nil {
+		multierror.Append(allErrors, err)
+	} else if len(result.Errors()) > 0 {
```
craigatgoogle: There's a potential NPE here. gojsonschema.Validate() has a return signature of (*Result, error), and returns nil for the result when it encounters an error while loading the schema. Calling result.Errors() in such a case will cause a NPE. I suggest returning the err from Validate() directly in your if block above to avoid this problem.

pkg/broker/variables_test.go
```diff
@@ -80,3 +82,177 @@ func TestBrokerVariable_ToSchema(t *testing.T) {
 		})
 	}
 }
+
+func TestBrokerVariable_ValidateVariables(t *testing.T) {
```
craigatgoogle: I suggest adding a case for when the schema syntax is incorrect to ensure that error is correctly handled.

mattysweeps: Added

pkg/broker/variables.go
mattysweeps: Updated

pkg/broker/variables.go
mattysweeps: Refactored to make this a bit cleaner. The NPE problem should be resolved.

PR #310, opened 2018-10-17
CloudSQL schemas

josephlewis42:
I'd like your opinion(s) on this. I've gone and defined the schemas before changing the code for `broker.go` so you can compare the two.

I added the ability to do assertions in the broker variables. The set variables won't affect anything when run, but we do need them in order to maintain the same functionality we had before.

If the CloudSQL API behaves correctly, then we shouldn't need the assertion at all.

Based on all the transformation I've been doing, I wrote down guidelines for future services we (and others) build out.

There is a "breaking" change in this code when combined with the JSONSchema code. The hard drive constants we were provisioning with before weren't in any way valid, I suspect the CloudSQL API was seeing an invalid value and defaulting back to SSD, but if that's the case then users that entered "hdd" thinking it was the other option would instead be getting the more expensive SSD as well. Therefore I'm considering it a bug fix instead.

I also have some hesitations that we're not doing PostgreSQL HA correctly, but that's another story.

### DIFF
--- a/brokerapi/brokers/cloudsql/common-definition.go
+++ b/brokerapi/brokers/cloudsql/common-definition.go
@@ -21,6 +21,12 @@ import (
 	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/varcontext"
 )
 
+const (
+	passwordTemplate   = "${rand.base64(32)}"
+	usernameTemplate   = `sb${str.truncate(14, time.nano())}`
+	identifierTemplate = `pcf-sb-${counter.next()}-${time.nano()}`
+)
+
 func roleWhitelist() []string {
 	return []string{
 		"cloudsql.editor",
@@ -59,8 +65,8 @@ func commonBindVariables() []broker.BrokerVariable {
 func commonBindComputedVariables() []varcontext.DefaultVariable {
 	return []varcontext.DefaultVariable{
 		// legacy behavior dictates that empty values get defaults
-		{Name: "password", Default: `${password == "" ? rand.base64(32) : password}`, Overwrite: true},
-		{Name: "username", Default: `${username == "" ? "sb${str.truncate(14, time.nano())}" : username}`, Overwrite: true},
+		{Name: "password", Default: `${password == "" ? "` + passwordTemplate + `" : password}`, Overwrite: true},
+		{Name: "username", Default: `${username == "" ? "` + usernameTemplate + `" : username}`, Overwrite: true},
 	}
 }
 
@@ -117,7 +123,7 @@ func commonProvisionVariables() []broker.BrokerVariable {
 		{
 			FieldName: "maintenance_window_day",
 			Type:      broker.JsonTypeString,
-			Details:   "(only for 2nd generation instances) This specifies when a v2 CloudSQL instance should preferably be restarted for system maintenance puruposes. Day of week (1-7), starting on Monday.",
+			Details:   "(only for 2nd generation instances) This specifies when a v2 CloudSQL instance should preferably be restarted for system maintenance purposes. Day of week (1-7), starting on Monday.",
 			Default:   "1",
 			Enum: map[interface{}]string{
 				"1": "Monday",
@@ -189,10 +195,10 @@ func commonProvisionVariables() []broker.BrokerVariable {
 func commonBindOutputVariables() []broker.BrokerVariable {
 	return []broker.BrokerVariable{
 		// Service account credentials (Note: they're a subset of the service account fields returned in a normal request)
-		{FieldName: "Email", Type: broker.JsonTypeString, Details: "Email address of the service account"},
+		{FieldName: "Email", Type: broker.JsonTypeString, Details: "Email address of the service account."},
 		{FieldName: "PrivateKeyData", Type: broker.JsonTypeString, Details: "Service account private key data. Base-64 encoded JSON."},
-		{FieldName: "ProjectId", Type: broker.JsonTypeString, Details: "ID of the project that owns the service account"},
-		{FieldName: "UniqueId", Type: broker.JsonTypeString, Details: "Unique and stable id of the service account"},
+		{FieldName: "ProjectId", Type: broker.JsonTypeString, Details: "ID of the project that owns the service account."},
+		{FieldName: "UniqueId", Type: broker.JsonTypeString, Details: "Unique and stable id of the service account."},
 
 		// Certificate
 		{FieldName: "CaCert", Type: broker.JsonTypeString, Details: "The server Certificate Authority's certificate."},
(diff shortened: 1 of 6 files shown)


brokerapi/brokers/cloudsql/mysql-definition.go
```diff
@@ -173,10 +179,70 @@ func init() {
...
+				Constraints: validation.NewConstraintBuilder().
+					Pattern("^[a-z][a-z0-9-]+$").
+					MaxLength(84).
+					Build(),
+			},
+			{
+				FieldName: "database_name",
+				Type:      broker.JsonTypeString,
+				Details:   "Name of the database inside of the instance. Must be a valid identifier for your chosen database type.",
+				Default:   "pcf-sb-${counter.next()}-${time.nano()}",
```
josephlewis42: I'm leaving the validation for these blank because this is implementation specific.

brokerapi/brokers/cloudsql/postgres-definition.go
```diff
@@ -128,10 +134,68 @@ func init() {
...
+				FieldName: "version",
+				Type:      broker.JsonTypeString,
+				Details:   "The database engine type and version.",
+				Default:   "POSTGRES_9_6",
+				Enum: map[interface{}]string{
+					"POSTGRES_9_6": "PostgreSQL 9.6.X",
+				},
+			},
+			{
+				FieldName: "failover_replica_name",
```
josephlewis42: Based on the docs I'm not convinced that this is the correct way to create a HA Postgres instance.

brokerapi/brokers/cloudsql/common-definition.go
```diff
@@ -17,167 +17,198 @@ package cloudsql
...
+		// legacy behavior dictates that empty values get defaults
+		{Name: "password", Default: `${password == "" ? rand.base64(32) : password}`, Overwrite: true},
+		{Name: "username", Default: `${username == "" ? "sb${str.truncate(14, time.nano())}" : username}`, Overwrite: true},
+	}
+}
+
+func commonProvisionVariables() []broker.BrokerVariable {
+	return []broker.BrokerVariable{
+		{
+			FieldName: "binlog",
```
josephlewis42: I'm not convinced that binlog is a valid property for Postgres instances.

brokerapi/brokers/cloudsql/common-definition.go
```diff
@@ -17,167 +17,198 @@ package cloudsql
...
-		Details:   "The activation policy specifies when the instance is activated; it is applicable only when the instance state is RUNNABLE.",
-		Default:   "ON_DEMAND",
-		Enum: map[interface{}]string{
-			"ALWAYS":    "Always, instance is always on.",
-			"NEVER":     "Never, instance does not turn on if a request arrives.",
-			"ON_DEMAND": "On Demand, instance responses to incoming requests and turns off when not in use.",
+func commonBindComputedVariables() []varcontext.DefaultVariable {
+	return []varcontext.DefaultVariable{
+		// legacy behavior dictates that empty values get defaults
+		{Name: "password", Default: `${password == "" ? rand.base64(32) : password}`, Overwrite: true},
```
craigatgoogle: Instead of repeating your config macro strings here, consider defining them as constants.

brokerapi/brokers/cloudsql/common-definition.go
```diff
@@ -17,167 +17,198 @@ package cloudsql
...
+			Default:   "PD_SSD",
+			Enum: map[interface{}]string{
+				"PD_SSD": "flash storage drive",
+				"PD_HDD": "magnetic hard drive",
+			},
+		},
+		{
+			FieldName: "maintenance_window_day",
+			Type:      broker.JsonTypeString,
+			Details:   "(only for 2nd generation instances) This specifies when a v2 CloudSQL instance should preferably be restarted for system maintenance puruposes. Day of week (1-7), starting on Monday.",
```
craigatgoogle: s/puruposes/purposes

docs/custom-services.md
```diff
@@ -42,3 +50,66 @@ The following string interpolation functions are available for use:
...
+
+For example, a Pub/Sub instance with one-to-many semantics might default to a read-only role, assuming the default consumer is just going to be a worker node whereas a Pub/Sub instance with many-to-many semantics might default to a read/write role even if some consumers want to be read-only.
+
+### Deciding on what your default plans will look like
+
+If you've gotten to this point, you should have a clear understanding of what your service is trying to accomplish, who the users are, and what variables are configurable in your plans.
+It can be tempting to include every permutation of the variables for plans.
+However, less is more.
+Operators need to look at each plan, decide if it fits a distinct use-case, budget and security model then make it available to individual teams.
+A few plans that hit key use-cases are much easier to grok.
```
craigatgoogle: Love the Heinlein reference ;)

brokerapi/brokers/cloudsql/common-definition.go
josephlewis42: Aww, man! That's a direct copy/paste from the docs. It looks like it rolls all the way back up through the client libs to the proto. Guess I'll be filing some bugs today.

docs/custom-services.md
```diff
@@ -42,3 +50,66 @@ The following string interpolation functions are available for use:
...
+
+* The users, developers who will use your service to provision things they work with day-to-day.
+* The operators, the people who are responsible for approving services and plans for developers to use.
+* Yourself, the person who has to maintain the service, strike the right balance of power between the operators and users, and make sure and make sure the new plans/services work as intended.
+
+The following sections contain guidelines to help you out.
+
+### Deciding what to include
+
+Services don't need to map one-to-one with cloud products, and probably shouldn't.
```
mattysweeps: Could we reference CloudSQL as an example?

docs/custom-services.md
```diff
@@ -42,3 +50,66 @@ The following string interpolation functions are available for use:
...
+Ideally a service will allow you to get a single, useful, task done and use plans to decide what kind of scale you want it done at.
+
+If you find yourself wishing you could selectively enable or disable variables based on flags, it's a sign you should break down your code into another service.
+For example, a Cloud Storage bucket can be configured to have a retention policy, a public-facing URL, and/or push file-change updates to a Pub/Sub queue.
+It would be a good idea to break those features into two separate services.
+One for hosting a static website with settings for URL, index/error pages, and CNAME.
```
mattysweeps: Bullet list these two?

docs/custom-services.md
```diff
@@ -42,3 +50,66 @@ The following string interpolation functions are available for use:
...
+And the other for general storage that has retention policies.
+Both could make use of a single variable for a queue.
+This helps while creating plans for them too.
```
mattysweeps: This feels like the start of a new paragraph

PR #311, opened 2018-10-17
migrated CloudSQL to BrokerBase

josephlewis42:
This change simplifies the way we handle brokers and account management.

Now every broker extends `broker_base` and all the CloudSQL specific code is contained under one directory. Refactoring after this merge can be done to integrate the account management with the rest of the bind process and to re-use pieces that were copied/pasted between the account manager and broker (like client creation and polling).

This refactoring passed all integration and unit tests. 

Helps with #305 and #250


brokerapi/brokers/cloudsql/broker.go
```diff
@@ -417,18 +417,18 @@ func combineServiceBindingCreds(sqlCreds models.ServiceBindingCredentials, saCre
...
-	return b.AccountManager.BuildInstanceCredentials(ctx, bindDetails, instanceDetails)
+	return b.sqlAccountManager().BuildInstanceCredentials(ctx, bindDetails, instanceDetails)
 }
 
 // Unbind deletes the database user, service account and invalidates the ssl certs associated with this binding.
 func (b *CloudSQLBroker) Unbind(ctx context.Context, creds models.ServiceBindingCredentials) error {
 	var accumulator error
 
-	if err := b.AccountManager.DeleteCredentials(ctx, creds); err != nil {
+	if err := b.sqlAccountManager().DeleteCredentials(ctx, creds); err != nil {
```
craigatgoogle: Instead of having separate account manager objects and calling DeleteCreds() on both of them, I wonder if it might not be appropriate to take an approach where SqlAccountManager extends AccountManager, overriding DeleteCreds() and calling super.DeleteCreds() OO internally traditional OO style. It might clean up this upstream code a bit. Here's a good article which shows an example of how to do this in Go: https://medium.com/random-go-tips/method-overriding-680cfd51ce40

brokerapi/brokers/cloudsql/sql_account_manager.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	if err := broker.pollOperationUntilDone(ctx, op, broker.ProjectId); err != nil {
+		return fmt.Errorf("Error encountered waiting for operation %q to finish: %s", op.Name, err)
+	}
+
+	return nil
+}
+
+func toJsonMap(obj interface{}) (map[string]interface{}, error) {
```
craigatgoogle: It seems like toJsonMap and mergeJsonProperties are orthogonal to the SQL broker's responsibilities. Consider moving them to a separate library file and adding unit tests so that other parts of the system can benefit from their utility.

PR #312, opened 2018-10-19
fixes for the JSONSchema checker that failed tests

josephlewis42:
Three changes I had to make when testing the JSONSchema stuff on provision. We should get Travis to run `go test ./...` on PRs as a sanity check.

* instanceID -> serviceId
* if raw params is nil, treat it as an empty map
* the defaults were causing regex checks to fail when the defaults were computed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: go
+go: "1.10"
+
--- a/brokerapi/brokers/brokers_test.go
+++ b/brokerapi/brokers/brokers_test.go
@@ -16,7 +16,6 @@ package brokers_test
 
 import (
 	"context"
-	"encoding/json"
 	"os"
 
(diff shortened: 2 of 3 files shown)


mattysweeps:
Whoops...

I'll add a commit to this PR to add TravisCI

mattysweeps:
The failing test is flaky. 
https://github.com/GoogleCloudPlatform/gcp-service-broker/blob/schema-fix/brokerapi/brokers/brokers_test.go#L500

Sometimes the db insert either doesn't work or hasn't finished by the time the bind call is invoked.

josephlewis42:
I think we should throw those tests out entirely. They're testing with specific implementations of ServiceBrokerHelper rather than a fake and keep breaking because they're testing implementation over behavior.

Since all the other tests are passing, and gcp_service_broker is going to be rewritten, including new tests, as part of #250 I think it's okay to merge with this failure, especially with it being transient. WDYT?

PR #313, opened 2018-10-19
Validate bind input parameters

mattysweeps:
* add feature flag for input validation

I tacked on the boolean to the `GCPServiceBroker` and `BrokerConfig` structs. I'm open to alternative designs, as I'm new to these woods.

Also regenerated all the things

### DIFF
--- a/brokerapi/brokers/config/broker_config.go
+++ b/brokerapi/brokers/config/broker_config.go
@@ -22,9 +22,10 @@ import (
 )
 
 type BrokerConfig struct {
-	Catalog    map[string]models.Service
-	HttpConfig *jwt.Config
-	ProjectId  string
+	Catalog               map[string]models.Service
+	HttpConfig            *jwt.Config
+	ProjectId             string
+	EnableInputValidation bool
 }
 
 func NewBrokerConfigFromEnv() (*BrokerConfig, error) {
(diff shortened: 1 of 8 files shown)


pkg/generator/forms.go
```diff
@@ -229,6 +230,24 @@ func generateCompatibilityForm() Form {
 	}
 }
 
+func generateValidationForm() Form {
```
josephlewis42: WDYT about turning this into the "flags" form as the home for various flags in the future?

josephlewis42: This flag could alsoo potentially live in the compatibility form since it'll be force-enabled in v5.0.

brokerapi/brokers/config/broker_config.go
```diff
@@ -22,9 +22,10 @@ import (
...
-	Catalog    map[string]models.Service
-	HttpConfig *jwt.Config
-	ProjectId  string
+	Catalog               map[string]models.Service
+	HttpConfig            *jwt.Config
+	ProjectId             string
+	EnableInputValidation bool
 }
 
 func NewBrokerConfigFromEnv() (*BrokerConfig, error) {
```
josephlewis42: I think the EnableInptuValidation flag should be checked and set in this function.

pkg/generator/forms.go
mattysweeps: I thought it should exist in compatibility as well. I'll but it there

PR #314, opened 2018-10-19
add tests for the oddities around CloudSQL before refactoring

josephlewis42:
These tests are temporary and will ensure that the requests generated before and after the move to `varcontext` are the same.

### DIFF
--- a/brokerapi/brokers/cloudsql/broker.go
+++ b/brokerapi/brokers/cloudsql/broker.go
@@ -63,16 +63,44 @@ type InstanceInformation struct {
 
 // Provision creates a new CloudSQL instance from the settings in the user-provided details and service plan.
 func (b *CloudSQLBroker) Provision(ctx context.Context, instanceId string, details brokerapi.ProvisionDetails, plan models.ServicePlan) (models.ServiceInstanceDetails, error) {
-	// validate parameters
-	var params map[string]string
-	var err error
+	di, ii, err := createProvisionRequest(instanceId, details, plan)
 
-	// validate parameters
+	// init sqladmin service
+	sqlService, err := b.createClient(ctx)
+	if err != nil {
+		return models.ServiceInstanceDetails{}, err
+	}
+
+	// make insert request
+	op, err := sqlService.Instances.Insert(b.ProjectId, &di).Do()
+	if err != nil {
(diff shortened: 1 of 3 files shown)


brokerapi/brokers/cloudsql/broker.go
```diff
@@ -63,16 +63,44 @@ type InstanceInformation struct {
 
 // Provision creates a new CloudSQL instance from the settings in the user-provided details and service plan.
 func (b *CloudSQLBroker) Provision(ctx context.Context, instanceId string, details brokerapi.ProvisionDetails, plan models.ServicePlan) (models.ServiceInstanceDetails, error) {
-	// validate parameters
-	var params map[string]string
-	var err error
+	di, ii, err := createProvisionRequest(instanceId, details, plan)
```
mattysweeps: Should this error be checked?

brokerapi/brokers/cloudsql/broker.go
```diff
@@ -63,16 +63,44 @@ type InstanceInformation struct {
...
+}
+
+func createProvisionRequest(instanceId string, details brokerapi.ProvisionDetails, plan models.ServicePlan) (di googlecloudsql.DatabaseInstance, ii InstanceInformation, err error) {
+	var params map[string]string
 
 	if len(details.RawParameters) == 0 {
 		params = map[string]string{}
 	} else if err = json.Unmarshal(details.RawParameters, &params); err != nil {
-		return models.ServiceInstanceDetails{}, fmt.Errorf("Error unmarshalling parameters: %s", err)
+		return di, ii, fmt.Errorf("Error unmarshalling parameters: %s", err)
```
mattysweeps: any reason for passing these pack on error, vs nil? (not a leading question. I honestly wouldn't know which is more idiomatic)

Update: at this point you're passing back default `di`  and `ii` interfaces. Which is different from `nil`, but still: if an error is thrown, when would they be safe to use? If you make all the error exits send nil, you could also remove the need for naming the return parameters in the first place.

josephlewis42: my goal was to simplify the refactoring without adding too much magic to the existing code, but I'll pull them out.

PR #317, opened 2018-10-22
added the concept of constants to varcontext which can't be overridden

josephlewis42:
This PR adds a concept of constants which aren't overridable by users or available in the output, they're simply there to be used if required.

These could include, for example, project ID, instance ID, service ID, default labels, instance information and mathematical constants.

This isn't the most elegant solution for a DSL, but a more complicated one would likely be overkill because this is the last major use-case for our internal "language" to make is suitable converting our existing variables over.

### DIFF
--- a/docs/custom-services.md
+++ b/docs/custom-services.md
@@ -51,6 +51,18 @@ The following string interpolation functions are available for use:
   * Generates `count` bytes of cryptographically secure randomness and converts it to [URL Encoded Base64](https://tools.ietf.org/html/rfc4648).
   * The randomness makes it suitable for using as passwords.
 
+## Variable reference
+
+The broker makes additional variables available to be used during provision and bind calls.
+
+### Provision
+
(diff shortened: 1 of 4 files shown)


mattysweeps:
@josephlewis42 Can you update this PR to use the constants section in a service definition?
If I could see a service using the constant map, it would be helpful for me to see the whole picture.

josephlewis42:
@mattysweeps I added some basic constants to the provision call. None of the services use them yet, but I think you might be able to see how they would.

In bind calls, there will be an additional namespace `instance.*` that will hold info about the existing instance. Probably the following fields from the `ServiceInstanceDetailsV1` struct:

```
	ID        string `gorm:"primary_key;type:varchar(255);not null"`
	Name         string
	OtherDetails string `gorm:"type:text"`
```

As:

```
instance.id -> string
instance.name -> string
instance.details -> map[string]interface{}
```

PR #324, opened 2018-10-24
move to varcontext on bind

### LINKED ISSUE #98: Configure defaults for provisioning request parameters

### LINKED ISSUE #304: Use Varcontext on Bind
Convert our bind requests to use varcontext.

josephlewis42:
This PR moves us to using `varcontext` on bind by constructing the context then passing it through.

Services with fixed accounts now get an entry in their `BindComputedVariables` to include those as fixed values.

Fixes #304 
Fixes #98


brokerapi/brokers/account_managers/service_account_manager.go
```diff
@@ -37,53 +36,41 @@ import (
...
+func (sam *ServiceAccountManager) CreateCredentials(ctx context.Context, vc *varcontext.VarContext) (map[string]interface{}, error) {
+	role := vc.GetString("role")
+	accountId := vc.GetString("service_account_name")
+	displayName := vc.GetString("service_account_display_name")
 
-func extractRole(details brokerapi.BindDetails) (string, error) {
-	bindParameters := map[string]interface{}{}
-	if err := json.Unmarshal(details.RawParameters, &bindParameters); err != nil {
-		return "", err
+	if err := vc.Error(); err != nil {
```
evandbrown: Use of `go-multierror` really enhances code readability. Nice!

PR #333, opened 2018-10-31
Autogenerate PCF forms

josephlewis42:
Auto-generate the `tile.yml` and `manifest.yml` files and added release notes about steps to take to generate them.

Until we move to using docker we'll need to continue to use the hard-coded version number because we don't have control over the flags that the buildpack uses to build the application.


josephlewis42:
edit: this merges into master so the version is 4.0.0, when I branch for 4.1, I'll follow the steps in the docs and bump/re-generate for that release branch.

pkg/generator/pcf-artifacts.go
```diff
@@ -0,0 +1,141 @@
...
+
+---`
+	manifestYmlTemplate = copyrightHeader + `
+applications:
+  - name: {{.appName}}
+    product_version: "{{.appVersion}}"
+    metadata_version: "1.0"
+    label: 'GCP Service Broker'
+    description: '{{.appDescription}}'
+    memory: 1G
```
mattysweeps: Any reason why these are not const values as well?
Even though they won't be changed often, if ever, still might be a good idea for consistency.

josephlewis42: I originally tried to just pull out the common vars, but I like your way better.

Issue #336, opened 2018-11-01
New tags for service providers


josephlewis42:
All of our service providers should get new tags from two pools:

Lifecycle:

These represent state of the service in GCP:

* ga
* beta
* deprecated

Broker:

These represent the state of the service in the broker:

* preview - this service is new to the broker and might not be complete, after 1 successful release this tag will be removed
* deprecated - this service is deprecated in the broker

NOTE: there is no distinction between a deprecated GCP and broker API because users need to transition either way.

These will be enabled using some new toggles to help operators quickly setup their broker:

* enable-ga-services (default true)
* enable-beta-services (default true)
* enable-deprecated-services (default false)
* enable-preview-services (default false)

All new Terraform provided resources will be in the `preview` state and therefore disabled by default.

PR #337, opened 2018-11-01
added toggles framework

josephlewis42:
Add a toggles framework that works on top of Viper and with our documentation generation system to generate docs.

This will help standardize the new toggles we need to add in 4.2 and beyond to support Terraform. See issue #336 for more info about the new service lifecycle management proposal.


docs/customization.md
```diff
@@ -355,25 +355,25 @@ Legacy Compatibility Options
...
+<b><tt>GSB_COMPATIBILITY_ENABLE_INPUT_VALIDATION</tt></b> - <i>boolean</i> - enable-input-validation
 
-Enable compatibility with the GCP Service Broker v3.x. Before version 4.0, each installation generated its own plan UUIDs, after 4.0 they have been standardized. This option installs a compatibility layer which checks if a service is using the correct plan GUID. If the service does not use the correct GUID, the request will fail with a message about how to upgrade.
+Enables validating user input variables against JSON Schema definitions.
 
 
 
 <ul>
   <li><b>Required</b></li>
+  <li>Default: <code>true</code></li>
```
craigatgoogle: If the field is required, does it make sense for it to have a default?

pkg/toggles/toggle.go
```diff
@@ -0,0 +1,104 @@
...
+	Description    string
+	propertyPrefix string
+}
+
+// EnvironmentVariable gets the environment variable used to control the toggle.
+func (toggle Toggle) EnvironmentVariable() string {
+	return utils.PropertyToEnv(toggle.viperProperty())
+}
+
+// viperProperty gets the viper property of the variable to contorl the toggle.
```
craigatgoogle: s/contorl/control

pkg/toggles/toggle_test.go
```diff
@@ -0,0 +1,59 @@
...
+// limitations under the License.
+
+package toggles
+
+import (
+	"fmt"
+
+	"github.com/spf13/viper"
+)
+
```
craigatgoogle: The examples are helpful, however there doesn't seem to be any unit tests verifying the logic within the toggles package.

docs/customization.md
josephlewis42: good catch!

pkg/toggles/toggle_test.go
josephlewis42: `go test` will validate the examples against the Output blocks at the end, I'm also happy to add other tests if you think they're warranted. I started with the examples because the tests seemed to fit that style better than the more declarative table driven one.

craigatgoogle: I forgot about that feature, which I'm sure you've mentioned before. SGTM, again ;)

PR #338, opened 2018-11-01
Make services shareable in Cloud Foundry

skibum55:
This PR adds the metadata to allow [shareable ](https://github.com/openservicebrokerapi/servicebroker/blob/v2.14/profile.md#cloud-foundry-service-metadata) services in Cloud Foundry.


josephlewis42:
Hi @skibum55,

Would you mind providing a little more context around this PR? Are there any privacy/security concerns about enabling it for all services or is there any reason it would cause a breaking change for existing users or why a platform operator would not want it on by default?

Thanks!
- Joseph

skibum55:
Enabling sharing for all services would introduce no additional security concerns.  The Platform Operator can disable sharing across the entire deployment using a feature flag and the service instance owner would need to explicitly share the instance with another CF org or space.  Services which used previous versions of the broker to create their instances would be unable to share, so no breaking changes there.    Sean

evandbrown:
@skibum55 I count 4 security considerations called out in the Pivotal docs at https://docs.pivotal.io/pivotalcf/2-1/services/enable-sharing.html#security

I understand a customer opts-in to this feature, but do they understand that there is nothing in place to prevent the org/space a service is shared with from escalating their privileges on bind? That seems like a big problem.

josephlewis42:
Thanks @mattmcneeney for the notes. I don't think I read anywhere that CF enforced the developer needing `SpaceDeveloper` in both source and destination spaces. That re-balances the equation if there's mutual trust between the developers of both spaces.

I've implemented this slightly differently in a separate PR as a flag that can be enabled in the tile so an operator can set the shareable flag on all services provided by the broker. That way operators can opt-in, K8s users are unaffected, and we don't need a flag per-service. That change should make it out in the 4.2 release in the first week of the new year.

PR #355, opened 2018-11-13
Terraform Support

josephlewis42:
This PR adds support for Terraform backends. Initially there is a single hard-coded service that must be explicitly enabled using the `enable-terraform-services` flag (off by default).

The new back-end works by creating an adapter to the exiting broker registry. The adapter creates Terraform "workspaces" which contain the tfstate, module definition, environment variables (ephemeral), and module instantiation variables. These workspaces can be serialized/deserialized between their state on disk to a database entry.

These workspaces support the `validate`, `provision`, and `deprovision` operations.   When a long-running operation is executed, the broker starts a background Terraform process, and updates the database with the operation (provision|deprovision) and returns a GUID that can be polled by the OSB client as part of the spec. When the operation completes, the database record is updated with the current tfstate and result of the operation.

Terraform output data is read directly from the consumed JSON tfstate file rather than calling the terraform binary and scraping the output. This is a calculated risk as Hashicorp could choose to change the format of the tfstate file, but we will be able to explicitly decide which versions of Terraform the broker is bundled with _and_ changes should be caught with our integration tests suddenly being unable to parse the file. It's also a better approach than trying to parse the CLI and recover things like types which are explicit in the JSON file.

The majority of this code is validated using our integration tests because it lies directly on the execution path.

Risks/deficiencies to be mitigated in future commits/releases:

* This code DOES NOT handle broker crashes. 
* This code DOES NOT handle concurrent access by multiple instances of the broker.
* This code DOES NOT prevent Terraform from downloading newer versions of the GCP plugins.
* This code does a straight transformation of the varcontext into a map, meaning the types of all computed variables will be strings. This prevents things like labels from being used.
* This code does not support loading custom service definitions.
* Testing can be improved if we start supporting things like custom paths for the terraform binary (which we can then mock out).


db_service/dao.go
```diff
@@ -360,3 +360,60 @@ func (ds *SqlDatastore) CheckDeletedPlanDetailsV1ById(ctx context.Context, id st
 }
```
josephlewis42: autogenerated

db_service/dao_test.go
```diff
@@ -35,6 +35,7 @@ func newInMemoryDatastore(t *testing.T) *SqlDatastore {
 	testDb.CreateTable(models.ServiceBindingCredentials{})
 	testDb.CreateTable(models.ProvisionRequestDetails{})
 	testDb.CreateTable(models.PlanDetailsV1{})
+	testDb.CreateTable(models.TerraformDeployment{})
```
josephlewis42: autogenerated file

pkg/providers/tf/definition.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil, err
+	}
+
+	return &broker.ServiceDefinition{
+		Name:                     tfb.Name,
+		DefaultServiceDefinition: string(defaultServiceDefinition),
+		ProvisionInputVariables:  tfb.ProvisionSettings.UserInputs,
+		ProvisionComputedVariables: append(tfb.ProvisionSettings.Computed, varcontext.DefaultVariable{
+			Name:      "tf_id",
+			Default:   "tf:${request.instance_id}:",
```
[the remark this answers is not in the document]
josephlewis42: These get processed in the GcpServiceBroker code which handles incoming HTTP connections. It grabs the user input, passes it all to the varcontext engine which executes the defaults and produces a varcontext. There's some additional context about what variables are available in the [variable reference](https://github.com/GoogleCloudPlatform/gcp-service-broker/blob/master/docs/custom-services.md#variable-reference) section of the docs.

pkg/providers/tf/job_runner.go
```diff
@@ -0,0 +1,233 @@
...
+
+	workspaceString, err := workspace.Serialize()
+	if err != nil {
+		return err
+	}
+
+	deployment := &models.TerraformDeployment{
+		ID:                jobId,
+		Workspace:         workspaceString,
+		LastOperationType: "validation",
```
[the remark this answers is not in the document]
josephlewis42: This is informational, usually a Provision will be issued right after this, but if it doesn't (like if the operator killed the process) then this state will be in the DB. Nothing will be returned to the user between this call and that one, so the cloud-controller would retry with a new GUID or tell the user to retry.

I went with this approach because I always wanted workspaces to be in a valid state in the DB because it made the preconditions for Create and Destroy more consistent. In my original approach, I didn't commit anything to the DB until the job had started, but that made the Create and Destroy methods fundamentally different, when Terraform really treats them the same which felt wrong.

pkg/providers/tf/job_runner.go
```diff
@@ -0,0 +1,233 @@
...
+func (runner *TfJobRunner) operationFinished(err error, workspace *wrapper.TerraformWorkspace, deployment *models.TerraformDeployment) error {
+	if err == nil {
+		deployment.LastOperationState = Succeeded
+		deployment.LastOperationMessage = ""
+	} else {
+		deployment.LastOperationState = Failed
+		deployment.LastOperationMessage = err.Error()
+	}
+
+	workspaceString, err := workspace.Serialize()
```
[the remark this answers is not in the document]
josephlewis42: It'll actually be this function that fails if the output changes: https://github.com/GoogleCloudPlatform/gcp-service-broker/blob/terraform/pkg/providers/tf/wrapper/workspace.go#L215

It unmarshals into those structs at the bottom of the file which correspond to just enough of the tfstate to get what they need.

pkg/providers/tf/cloud-storage.go
```diff
@@ -0,0 +1,191 @@
...
+		"storage.objectViewer",
+		"storage.objectAdmin",
+	}
+
+	cloudStorage := TfServiceDefinitionV1{
+		Version:          1,
+		Name:             "google-storage-experimental",
+		Id:               "68d094ae-e727-4c14-af07-ee34133c8dfb",
+		Description:      "Unified object storage for developers and enterprises. Cloud Storage allows world-wide storage and retrieval of any amount of data at any time.",
+		DisplayName:      "Google Cloud Storage 2",
```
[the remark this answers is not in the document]
josephlewis42: That was a mistake, the name was originally `google-storage-2`, but I realized if an operator were to somehow enable this and get it working I wanted people to explicitly know it was experimental. I'll call it out here too.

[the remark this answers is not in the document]
josephlewis42: Updated the display name and the descrpition. The description calls out that it's experimental and binds the service account role directly to the bucket.

craigatgoogle:
Looping in @danawillow to this PR. AFAIK Hashicorp has been undergoing a major re-factoring of Terraform which is slated to be released sometime next year. I would want to check and see if this includes any changes to the TFState schema that should be accounted for in future planning. If there are, I would check to see if it's possible to get early access to the schema changes so we can have them ready to sim-ship when the new Terraform version launches.

pkg/providers/tf/provider.go
```diff
@@ -0,0 +1,172 @@
...
+	provider.logger.Info("terraform-deprovision", lager.Data{
+		"instance": instance.ID,
+	})
+
+	tfId := generateTfId(instance.ID, "")
+	if err := provider.jobRunner.Destroy(ctx, tfId); err != nil {
+		return nil, err
+	}
+
+	return &tfId, nil
```
mattysweeps: Why is there different behavior between the Unbind waiting and checking the result of the job runner's Destroy, and the Deprovision not waiting? 

Since both are terraform applys, it seems equally likely that the Deprovision could fail and leave behind infrastructure.

pkg/providers/tf/cloud-storage.go
```diff
@@ -0,0 +1,191 @@
...
+		"storage.objectCreator",
+		"storage.objectViewer",
+		"storage.objectAdmin",
+	}
+
+	cloudStorage := TfServiceDefinitionV1{
+		Version:          1,
+		Name:             "google-storage-experimental",
+		Id:               "68d094ae-e727-4c14-af07-ee34133c8dfb",
+		Description:      "Experimental Google Cloud Storage that uses the Terraform back-end and grants service accounts IAM directly on the bucket.",
```
craigatgoogle: s/accounts IAM/accounts IAM permissions

pkg/providers/tf/job_runner.go
```diff
@@ -0,0 +1,233 @@
...
+
+// Wait waits for an operation to complete, polling its status once per second.
+func (runner *TfJobRunner) Wait(ctx context.Context, id string) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+
+		case <-time.After(1 * time.Second):
+			isDone, err := runner.Status(ctx, id)
```
craigatgoogle: Will the runner time out on its own, or do we need to think about implementing that here to prevent infinite waiting?

danawillow:
@craigatgoogle yes, the format of the state file is changing, mostly around how nested objects are represented. If you try out the 0.12 alpha you should be able to see what the new format looks like. In the meantime, you should be able to pin the version of Terraform used so that you don't use 0.12 until you're ready for it.

pkg/providers/tf/job_runner.go
```diff
@@ -0,0 +1,233 @@
...
+	deployment.LastOperationMessage = ""
+
+	if err := db_service.SaveTerraformDeployment(ctx, deployment); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (runner *TfJobRunner) hydrateWorkspace(ctx context.Context, deployment *models.TerraformDeployment) (*wrapper.TerraformWorkspace, error) {
```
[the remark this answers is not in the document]
josephlewis42: I used hydrate because we aren't really creating the workspace on disk, nor are we really creating a new one, but we are deserializing it and augmenting it with additional data.

craigatgoogle:
Awesome, thanks @danawillow :)

pkg/providers/tf/cloud-storage.go
```diff
@@ -0,0 +1,191 @@
...
+						Pattern("^[A-Za-z][-a-z0-9A-Z]+$").
+						Examples("US", "EU", "southamerica-east1").
+						Build(),
+				},
+			},
+			Computed: []varcontext.DefaultVariable{
+				{Name: "labels", Default: "${json.marshal(request.default_labels)}", Overwrite: true},
+			},
+			Template: `
+	variable name {type = "string"}
```
craigatgoogle: Having blocks of in-line TF templates like this makes me concerned that as TF becomes the primary source of provisioning for the SB, these in-line templates end up being a growing hole in the test coverage for the project. Consider splitting the templates out and rigging up unit tests for each template which validates the syntax.

craigatgoogle: Discussed this off-line with @josephlewis42 and he clarified that these templates do get utilized/validated and during execution of the examples during integration testing.

josephlewis42: I have another PR into this branch for this issue #356; it validates that the HCL is syntactically valid (HCL 1.0), and that the inputs and outputs match up with what's expected based on what the plugin author has listed.

Outputs must be an exact match, while the listed inputs can be a superset (some inputs could be deprecated or used in computation, but never by the actual Terraform definition).

craigatgoogle:
Still LGTM.

PR #360, opened 2018-11-27
set default roles for legacy plans

josephlewis42:
Set default roles for builtin legacy plans. These are tricky because the operator can override them, so we have to be mindful of the case where the operator might not include the default role in their list.

### DIFF
--- a/brokerapi/brokers/account_managers/service_account_manager.go
+++ b/brokerapi/brokers/account_managers/service_account_manager.go
@@ -36,9 +36,14 @@ import (
 	iam "google.golang.org/api/iam/v1"
 )
 
-const roleResourcePrefix = "roles/"
-const saResourcePrefix = "serviceAccount:"
-const projectResourcePrefix = "projects/"
+const (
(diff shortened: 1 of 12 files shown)


brokerapi/brokers/account_managers/service_account_manager.go
```diff
@@ -202,23 +207,27 @@ type ServiceAccountInfo struct {
 	PrivateKeyData string `json:"PrivateKeyData"`
 }
 
-func ServiceAccountBindInputVariables(serviceName string, defaultWhitelist []string) []broker.BrokerVariable {
-	details := fmt.Sprintf(`The role for the account without the "roles/" prefix.
-		See: https://cloud.google.com/iam/docs/understanding-roles for more details.
-		Note: The default enumeration may be overridden by your operator.`)
-
+// ServiceAccountWhitelistWithDefault holds overridable whitelists with default values.
```
craigatgoogle: The name in this comment block doesn't seem to match up with the function itself.

PR #366, opened 2018-12-03
Brokerpaks

josephlewis42:
So, this needs a bit of a documentation, but the brokerpaks work. Work following this PR I know _needs_ to be done is the following:

* Better examples in the `google-brokers` directory (and probably a re-working of the dir).
* Documentation, both a theory of brokerpaks, and execution doc about brokerpaks in practice.
* Support loading brokerpaks from HTTP/Git/S3/GCS (this can be done using the Hashicorp lib go-getter we already imported.
* Add support for brokerpaks to the tile.
* Refactor plugin loading so you can disable "legacy" services.

There's a three-pronged testing approach.

1) unit-tests for most of the new utility libraries.
2) integration testing building a new lib using: `gcp-service-broker pak test`
3) e2e testing the execution path using `gcp-service-broker pak run-examples mypak.brokerpak`


cmd/pak.go
```diff
@@ -0,0 +1,175 @@
...
+When you're done, you can build the bundle which will download the sources,
+Terraform resources, and pack them together.
+
+	cd .. && gcp-service-broker pak build my-pak
+
+This will produce a pack:
+
+	my-pak.brokerpak
+
+You can validate the pack:
```
craigatgoogle: Is there a reason for this to be made into its own separate step and not wrapped up in the build process? For example, can someone make a change to a pak which would need validation without having to first execute the build?

cmd/pak.go
```diff
@@ -0,0 +1,175 @@
...
+
+Bundles include source code (for legal compliance), service definitions, and
+Terraform/provider binaries for multiple platforms. They give you a contained
+way to deploy new services to existing brokers or augment the broker to fit
+your needs.
+
+To start building a pack, create a new directory and within it run init:
+
+	mkdir my-pak && cd my-pak
+	gcp-service-broker pak init
```
craigatgoogle: nit: The build command allows you to specify a directory, where as init does not. I wonder if it makes sense for consistency sake to have init to accept an optional path arg or else default to the current directory.

cmd/pak.go
```diff
@@ -0,0 +1,175 @@
...
+
+You'll get a new pack with a manifest and example service definition.
+Define the architectures and Terraform plugins you need in your manifest along
+with any metadata you want, and include the names of all service definition
+files.
+
+When you're done, you can build the bundle which will download the sources,
+Terraform resources, and pack them together.
+
+	cd .. && gcp-service-broker pak build my-pak
```
craigatgoogle: Same comment as above, but conversely.

pkg/brokerpak/manifest.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (m *Manifest) packDefinitions(tmp, base string) error {
+	manifestCopy := *m
+
+	var servicePaths []string
```
mattysweeps: Can you leave a code comment explaining the reasoning for moving the definition files from their original location to the `definitions` directory?

pkg/brokerpak/reader.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (pak *BrokerPakReader) extractPlatformBins(destination string) error {
+	mf, err := pak.Manifest()
+	if err != nil {
+		return err
+	}
+
+	if !mf.AppliesToCurrentPlatform() {
+		return errors.New("The package does not contain the binaries necessary to run on the current platform.")
```
mattysweeps:
```suggestion
		return errors.New("the package does not contain the binaries necessary to run on the current platform.")
```

utils/ziputil/zip.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// Joins a path for use in a zip file.
+func Join(path ...string) string {
+	return strings.Join(path, "/")
+}
+
+// Clean converts a path into a zip style path by replacing backslashes with
+// forward-slashes and making the paths relative by removing leading "/" and "./".
+func Clean(path ...string) string {
```
mattysweeps: You might want to check for and error on the occurrence of `../`

(doesn't have to be this function)

cmd/pak.go
```diff
@@ -0,0 +1,175 @@
...
+		Short: "bundle up the service definition files and Terraform resources into a brokerpak",
+		Args:  cobra.MaximumNArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			directory := ""
+			if len(args) == 1 {
+				directory = args[0]
+			}
+
+			pakPath, err := brokerpak.Pack(directory)
+			if err != nil {
```
craigatgoogle: I'm curious why the Init() function seems to encapsulate its error reporting, whereas Pack() seems to have delegated it to here.

cmd/pak.go
```diff
@@ -0,0 +1,175 @@
...
+			return brokerpak.Info(args[0])
+		},
+	})
+
+	pakCmd.AddCommand(&cobra.Command{
+		Use:   "validate [pack.brokerpak]",
+		Short: "validate a brokerpak",
+		Args:  cobra.ExactArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := brokerpak.Validate(args[0]); err != nil {
```
craigatgoogle: Same question re: logging consistently as above.

google-brokers/manifest.yml
```diff
@@ -0,0 +1,19 @@
+packversion: 1
+name: google-services
+version: 1.0.0
+metadata:
+  author: jlewisiii@google.com
```
craigatgoogle: I would shy away from using my real email address in example tests, but your call :)

utils/ziputil/zip.go
```diff
@@ -0,0 +1,150 @@
...
+	"path/filepath"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/GoogleCloudPlatform/gcp-service-broker/utils/stream"
+)
+
+// List writes a ls -la style listing of the zipfile to the given writer.
+func List(z *zip.Reader, w io.Writer) {
+	sw := tabwriter.NewWriter(w, 0, 0, 2, ' ', tabwriter.StripEscape)
```
craigatgoogle: It's generally good practice when using literal arguments to functions like this to annotate them with a comment with the parameter name they correspond to. Makes for easier code reading.

pkg/brokerpak/cmd.go
```diff
@@ -0,0 +1,195 @@
...
+			fmt.Fprintf(iw, "\t%s\t%s\n", k, v)
+		}
+
+		iw.Flush()
+		fmt.Println()
+	}
+
+	{
+		fmt.Println("Dependencies")
+		rw := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', tabwriter.StripEscape)
```
craigatgoogle: This configuration of tabwriter seems to be repeated throughout. Consider making a factory which all of these locations all with this default config, in-case you want to change it later.

cmd/pak.go
josephlewis42: Ideally people won't go mucking about with the packs themselves, but you might download one off the internet and it might be incompatible with your version of the broker due to a schema change.

cmd/pak.go
josephlewis42: That was me abusing the Cobra library. RunE runs the command and if the error is returned it prints it and the `--help` page. RunE should really only be used if you're returning an invalid argument. I'll go fix that.

pkg/brokerpak/cmd.go
```diff
@@ -0,0 +1,195 @@
...
+
+	services, err := brokerPak.Services()
+	if err != nil {
+		return err
+	}
+
+	// Pack information
+	fmt.Println("Information")
+	{
+		iw := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', tabwriter.StripEscape)
```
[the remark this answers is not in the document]
josephlewis42: that was an artifact from an earlier version, I'll fix that up

pkg/brokerpak/cmd.go
```diff
@@ -0,0 +1,195 @@
...
+		fmt.Fprintln(sw, "ID\tNAME\tDESCRIPTION\tPLANS")
+		for _, svc := range services {
+			fmt.Fprintf(sw, "%s\t%s\t%s\t%d\n", svc.Id, svc.Name, svc.Description, len(svc.Plans))
+		}
+		sw.Flush()
+		fmt.Println()
+	}
+
+	fmt.Println("Contents")
+	ziputil.List(&brokerPak.contents.Reader, os.Stdout)
```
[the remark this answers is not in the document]
josephlewis42: There's not one on the list function.

pkg/brokerpak/fetch.go
```diff
@@ -0,0 +1,33 @@
...
+func fetchArchive(src, dest string) error {
+	return (&getter.Client{
+		Src:           src,
+		Dst:           dest,
+		Mode:          getter.ClientModeFile,
+		Getters:       getter.Getters,
+		Decompressors: map[string]getter.Decompressor{},
+	}).Get()
+
+	return getter.Get(src, dest)
```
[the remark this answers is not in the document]
josephlewis42: Nice catch!

pkg/brokerpak/fetch.go
```diff
@@ -0,0 +1,33 @@
...
+// fetchArchive uses go-getter to download archives. By default go-getter
+// decompresses archives, so this configuration prevents that.
+func fetchArchive(src, dest string) error {
+	return (&getter.Client{
+		Src:           src,
+		Dst:           dest,
+		Mode:          getter.ClientModeFile,
+		Getters:       getter.Getters,
+		Decompressors: map[string]getter.Decompressor{},
+	}).Get()
```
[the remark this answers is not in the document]
josephlewis42: getter will create parent directories in the destination which is something we take advantage of, but for src it would throw an error because it uses the src path to determine which handler to use (http/file/s3/git/etc.)

pkg/brokerpak/manifest.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/GoogleCloudPlatform/gcp-service-broker/utils/stream"
+	"github.com/GoogleCloudPlatform/gcp-service-broker/utils/ziputil"
+
+	getter "github.com/hashicorp/go-getter"
+)
+
+const manifestName = "manifest.yml"
+
+type Manifest struct {
+	PackVersion int `yaml:"packversion" validate:"required,eq=1"`
```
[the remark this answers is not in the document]
josephlewis42: I was trying to keep the stuff that the user modifies separate from the metadata.

cmd/pak.go
josephlewis42: ++

cmd/pak.go
josephlewis42: ++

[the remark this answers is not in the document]
josephlewis42: ++, as long as you're not running this in the root of the filesystem the app will be okay which I think is a reasonable risk.

pkg/brokerpak/platform_test.go
```diff
@@ -0,0 +1,39 @@
...
+	// Output: bsd/amd64
+}
+
+func ExamplePlatform_Equals() {
+	p := Platform{Os: "js", Arch: "webasm"}
+	fmt.Println(p.Equals(p))
+	fmt.Println(p.Equals(CurrentPlatform()))
+
+	// Output: true
+	// false
```
craigatgoogle: What is some lunatic happen to be running this test on js/webasm ;)

utils/stream/stream.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return fmt.Errorf("copy couldn't open destination: %v", err)
+	}
+	defer writeCloser.Close()
+
+	if _, err := io.Copy(writeCloser, readCloser); err != nil {
+		return fmt.Errorf("copy couldn't copy data: %v", err)
+	}
+
+	if err := readCloser.Close(); err != nil {
```
craigatgoogle: I'm concerned about the fact that Close() is called twice on both the writer and the reader. Here's what the docs have to say:
"The behavior of Close after the first call is undefined. Specific implementations may document their own behavior."

craigatgoogle: Just noticed the use of io.NoOpCloser below. I make make mention of that assumption in the method header for future maintainers.

utils/stream/stream.go
```diff
@@ -0,0 +1,189 @@
...
+// after the copy finishes.
+func ToWriteCloser(w io.WriteCloser) Dest {
+	return func() (io.WriteCloser, error) {
+		return w, nil
+	}
+}
+
+// bufferCallback buffers the results and on close calls the callback.
+type bufferCallback struct {
+	bytes.Buffer
```
craigatgoogle: Sorry for being dense, but where is the buffer being allocated for this. I'm used to seeing an accompanying NewBuffer() call.

utils/stream/stream_test.go
```diff
@@ -0,0 +1,125 @@
...
+		panic(err)
+	}
+
+	fmt.Printf("%#v", b)
+
+	// Output: stream.Test{Str:"foo", Num:42}
+}
+
+func ExampleFile() {
+	td, _ := ioutil.TempDir("", "test")
```
craigatgoogle: Probably don't want to swallow the error here even though its unlikely.

utils/stream/stream.go
craigatgoogle: Specifically I'm wondering about the case in stream_test.go, where bufferCallback{} is instantiated directly and then subsequently written to.

utils/ziputil/zip.go
```diff
@@ -0,0 +1,173 @@
...
+		fmt.Errorf("no such file: %q", Join(path...))
+	}
+
+	return f.Open()
+}
+
+// Extracts the contents of the zipDirectory to the given OS osDirectory.
+// This routine is overly strict and doesn't allow extracting _any_ files that
+// contain "..".
+func Extract(z *zip.Reader, zipDirectory, osDirectory string) error {
```
craigatgoogle: Not meaning to question the quality of the implementation here, but this all seems like very boiler-plate archive/extract stuff. Thinking towards minimizing the code we have to maintain, I'm curious what the state of prior art is regarding existing libraries in this space?

utils/ziputil/zip_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/gcp-service-broker/utils/stream"
+)
+
+var exampleFile = []byte{
```
craigatgoogle: Consider adding a comment describing the contents of the example file here.

utils/ziputil/zip_test.go
```diff
@@ -0,0 +1,148 @@
...
+func ExampleOpen() {
+	zf := openZipfile()
+	source := stream.FromReadCloserError(Open(zf, "path/to/my/file.txt"))
+	stream.Copy(source, stream.ToWriter(os.Stdout))
+
+	// Output: Hello, world!
+}
+
+func ExampleExtract() {
+	tmp, _ := ioutil.TempDir("", "ziptest")
```
craigatgoogle: Again recommend not swallowing the error here, instead consider creating a test factory which creates tempdirs and handles the error for you.

cmd/pak.go
craigatgoogle: That's a good point. Thanks.

utils/ziputil/zip.go
josephlewis42: I looked at several libraries trying to get around writing this (mostly so I wouldn't have to write the tests...). They all seem to be deficient in one way or another, not protecting against zip slip, not allowing unarchiving of a sub-directory (only the whole zip), forcing the file extension to be .zip (I opened a bug about that one, but the author won't accept a PR to change it because they think it's a feature), others are GPL, and some don't respect file permissions on extraction.

In an ideal world, I'd really like to use spf13's afero with plugins for zip and gcs: https://github.com/spf13/afero but the project looks dead at this point. Any thoughts about one you like and I might be able to use?

utils/ziputil/zip.go
josephlewis42: I added this further down using the same technique that Hashicorp used in go-getter, which was copied from the go stdlib.

pkg/brokerpak/platform_test.go
josephlewis42: You know I love all things JS and WASM, so that lunatic is likely to be me a year from now. :grin: Changing to beos/wasm, it was hard to choose that combo over templeos and haiku though...

utils/stream/stream.go
josephlewis42: You were right the first time, I'm adding a multi-closer to handle this case that will guarantee at-most-once close semantics.

I found lots of fun discussion about adding a `OnceCloser()` to the stdlib, but as always it ended up in the state of "this is only 10 lines, just add it to your own project"...ugh.

utils/stream/stream.go
josephlewis42: This one got me too, it turns out that the zero value of a buffer is the preferred way to initialize it in most cases: https://golang.org/src/bytes/buffer.go?s=14358:14392#L440

utils/ziputil/zip_test.go
josephlewis42: Preview may not be updating, I added this:

```go
// A zip file with a single file with the contents "Hello, world!":
// MODE        SIZE  NAME
// -rw-rw-rw-  13    path/to/my/file.txt

```

utils/ziputil/zip.go
craigatgoogle: I don't have any recommendations, just wanted to verify that all options were thoroughly explored. I trust your judgement here :)

utils/ziputil/zip.go
craigatgoogle: I noticed this feedback was resolved but I don't see any related changes. Any thoughts?

utils/stream/stream.go
craigatgoogle: Gotcha, thanks.

PR #373, opened 2018-12-06
added support for fetching brokerpaks from GCS

josephlewis42:
Adds support for getting brokerpaks from Cloud Storage buckets using URLs like gsutil produces e.g. `gs://brokerpaks/my-pak.brokerpak`

Testing manually works just fine, but I can't seem to come up with a good way to write unit-tests for the fetcher without massively increasing the code size and complexity due to the way the GCS client library is structured. This is similar to the issue I faced when creating the cloud storage beat for Elastic.

If you have thoughts beyond testing this with our integration tests I'm happy to hear them.

### DIFF
--- a/pkg/brokerpak/cmd.go
+++ b/pkg/brokerpak/cmd.go
@@ -159,6 +159,8 @@ func RegisterAll(registry broker.BrokerRegistry) error {
 	}
 	defer os.RemoveAll(pakDir)
 
+	// XXX(josephlewis42): this could be parallelized to increase performance
+	// if we find people are pulling lots of data from the network.
 	for i, packSource := range viper.GetStringSlice("brokerpak.packs") {
 		destFile := filepath.Join(pakDir, fmt.Sprintf("pack-%d.brokerpak", i))
 		registerLogger.Debug("importing brokerpak", lager.Data{
--- a/pkg/brokerpak/cmd.go
+++ b/pkg/brokerpak/cmd.go
@@ -16,14 +16,17 @@ package brokerpak
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"text/tabwriter"
 
+	"code.cloudfoundry.org/lager"
 	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/broker"
 	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/client"
 	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/generator"
 	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/providers/tf"
+	"github.com/GoogleCloudPlatform/gcp-service-broker/utils"
 	"github.com/GoogleCloudPlatform/gcp-service-broker/utils/stream"
 	"github.com/GoogleCloudPlatform/gcp-service-broker/utils/ziputil"
 	"github.com/spf13/viper"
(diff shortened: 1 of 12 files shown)


pkg/brokerpak/cmd.go
```diff
@@ -148,10 +151,30 @@ func Validate(pack string) error {
...
+	defer os.RemoveAll(pakDir)
+
+	for i, packSource := range viper.GetStringSlice("brokerpak.packs") {
+		destFile := filepath.Join(pakDir, fmt.Sprintf("pack-%d.brokerpak", i))
+		registerLogger.Debug("importing brokerpak", lager.Data{
+			"source":      packSource,
+			"destination": destFile,
+		})
+
+		if err := fetchBrokerpak(packSource, destFile); err != nil {
```
craigatgoogle: One potential optimization would be to implement a fan-out/fan-in parallel fetches here, but I wouldn't bother unless this actually becomes a bottleneck.

pkg/brokerpak/fetch.go
```diff
@@ -25,7 +40,91 @@ func fetchArchive(src, dest string) error {
...
+		return fmt.Errorf("couldn't turn dir %q into abs path: %v", execWd, err)
+	}
+
+	getters := map[string]getter.Getter{}
+	for k, g := range getter.Getters {
+		getters[k] = g
+	}
+	getters["gs"] = &gsGetter{}
+
+	return (&getter.Client{
```
craigatgoogle: This method seems to have two responsibilities: 1. initialize the client 2. fetch the broker pak. If you separated these concerns and had the client be passed in as an argument, it would improve orthogonality and allow for unit testing the logic of this module via mocking the client.

pkg/brokerpak/fetch.go
```diff
@@ -25,7 +40,91 @@ func fetchArchive(src, dest string) error {
...
+	if err != nil {
+		return err
+	}
+
+	src := stream.FromReadCloserError(obj.NewReader(ctx))
+	return stream.Copy(src, stream.ToFile(dst))
+}
+
+func (g *gsGetter) objectAt(u *url.URL) (*storage.ObjectHandle, error) {
+	client, err := g.client()
```
craigatgoogle: This method seems to have two responsibilities: 1. initialize the GCS client 2. fetch the object. If you separated these concerns and had the client be passed in to the constructor of the getter.Client , it would improve orthogonality and allow for unit testing the logic of this module via mocking the client.

mattysweeps: I agree that the `objectAt` method should accept a `*storage.Client` parameter.

mattysweeps: Since the only logic here is to trim the leading slash, I think this is fine to skip on unit testing and reply on integration testing only. Since to unit test you'd have to create a implementation of the storage.Client interface, which I agree is a lot of work for simply passing through a `gs://` url

pkg/brokerpak/fetch.go
```diff
@@ -25,7 +40,91 @@ func fetchArchive(src, dest string) error {
...
+func (g *gsGetter) GetFile(dst string, u *url.URL) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	defer cancel()
+
+	obj, err := g.objectAt(u)
+	if err != nil {
+		return err
+	}
+
+	src := stream.FromReadCloserError(obj.NewReader(ctx))
```
mattysweeps: I think the error on `obj.NewReader(ctx)` should be checked and raised, instead of silenced. Because of how the storage package is written, no "invalid bucket" or "invalid object" errors are thrown until the read call. So the src could be invalid (or empty, most likely) and `GetFile` might return no error.

pkg/brokerpak/cmd.go
josephlewis42: I added a note inline to keep track of this potential improvement.

josephlewis42:
I added tests around the assumptions we made about the upstream libraries we use and the behavior of the default client as well as de-duplicating some code.

PR #378, opened 2018-12-10
added osdfgen

josephlewis42:
This is a small utility I've been using to update the OSDF license based on `Gopkg.lock`. The intent is for it to only be run manually and never by end-users.

### DIFF
--- a/tools/osdfgen/osdfgen.go
+++ b/tools/osdfgen/osdfgen.go
@@ -1,0 +1,147 @@
+// Copyright 2018 the Service Broker Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"encoding/csv"
+	"errors"
+	"flag"
+	"fmt"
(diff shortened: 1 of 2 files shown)


tools/osdfgen/osdfgen.go
```diff
@@ -0,0 +1,147 @@
...
+
+	"gopkg.in/src-d/go-license-detector.v2/licensedb"
+	"gopkg.in/src-d/go-license-detector.v2/licensedb/filer"
+
+	toml "github.com/pelletier/go-toml"
+)
+
+var (
+	out  = flag.String("o", "-", "Sets the output location of the OSDF csv")
+	proj = flag.String("p", ".", "the project root")
```
craigatgoogle: Use of capitalization is inconsistent in the flag description strings.

tools/osdfgen/osdfgen.go
```diff
@@ -0,0 +1,147 @@
...
+type Lockfile struct {
+	Projects []Project
+}
+
+type Project struct {
+	Name     string
+	Revision string
+}
+
+func main() {
```
craigatgoogle: Consider adding a doc header explaining the usage either here or as a package header.

tools/osdfgen/osdfgen.go
```diff
@@ -0,0 +1,147 @@
...
+		if err != nil {
+			log.Printf("Could not get license text for %q: %s", project.Name, err)
+		}
+
+		// name, hash, spdx, full text
+		writer.Write([]string{project.Name, project.Revision, spdxCode, licenseText})
+	}
+
+	writer.Flush()
+
```
craigatgoogle: Extra '\n'

craigatgoogle:
Although a small utility may seem trivial at first glance, I'd still recommend implementing unit tests for the follow reasons: 1. They serve as a form of code documentation/example usage. 2. When some maintainer in the future who has no working knowledge of this code, (possibly yourself), needs to make a change it will help them to check for regressions.

PR #379, opened 2018-12-11
added naming recommendations

josephlewis42:


josephlewis42:
@craigatgoogle I like that idea, I think I can do it using the validation package still but with a different validation tag so you could `func Validate() error` and `func Warnings() error`. It would be fairly simple to implement `-Werror` like functionality then too.

PR #391, opened 2018-12-20
Fix for Postgres machine types

josephlewis42:
Moved all the Postgres machines to their equivalent custom machine types because PG only supports the standard machines if they have shared CPUs.

Added tests to make sure the values of the existing (and any new) custom machine types conform to the written spec.

### DIFF
--- a/brokerapi/brokers/cloudsql/broker_test.go
+++ b/brokerapi/brokers/cloudsql/broker_test.go
@@ -289,7 +289,6 @@ func TestPostgresCustomMachineTypes(t *testing.T) {
 
 	for _, plan := range sd.Plans {
 		t.Run(plan.Name, func(t *testing.T) {
-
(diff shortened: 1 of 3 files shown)


brokerapi/brokers/cloudsql/broker_test.go
```diff
@@ -279,3 +280,49 @@ func TestCreateProvisionRequest(t *testing.T) {
...
+
+func TestPostgresCustomMachineTypes(t *testing.T) {
+	sd, err := postgresServiceDefinition().ServiceDefinition()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, plan := range sd.Plans {
+		t.Run(plan.Name, func(t *testing.T) {
+
```
craigatgoogle: Extra '\n'

PR #392, opened 2018-12-20
Add support for JSONSchema in /catalog

### LINKED ISSUE #323: Return JSON Schemas during the /catalog call
Thanks to the work @mattysweeps has done, we can generate JSON schema on the fly. Next, we should add the definitions to the /v2/catalog endpoint (probably by modifying the BrokerService::CatalogEntry method). This will expose the correct schemas to the user through the CLI and supported web forms.

Correct meaning:

* up-to-date
* consistent with the configuration of this instance, including changes operators can make.

josephlewis42:
Fixes #323 

This PR adds support to export the JSONSchemas we use to validate user-input as part of our catalog call. This allows external clients to use that to render forms for users and do preemptive checking of the values users enter.

By default our forms don't provider user-friendly names to display in labels/tooltips, so I created a converter that can be overridden.

Here's an example for Redis:

```json
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "properties": {
      "authorized_network": {
        "default": "",
        "description": "The full name of the Google Compute Engine network to which the instance is connected. If left unspecified, the default network will be used. This will look something like /projects/PROJECTID/global/networks/NETWORKNAME.",
        "title": "Authorized Network",
        "type": "string"
      },
      "display_name": {
        "default": "${instance_id}",
        "description": "For display purposes only",
        "maxLength": 80,
        "title": "Display Name",
        "type": "string"
      },
      "instance_id": {
        "default": "pcf-sb-${counter.next()}-${time.nano()}",
        "description": "Permanent identifier for your instance",
        "maxLength": 30,
        "minLength": 6,
        "pattern": "^[a-z][a-z0-9-]+$",
        "title": "Instance ID",
        "type": "string"
      },
      "memory_size_gb": {
        "default": 4,
        "description": "The size of the instance in gigabytes.",
        "maximum": 300,
        "minimum": 1,
        "title": "Memory Size GB",
        "type": "integer"
      },
      "region": {
        "default": "us-central1",
        "description": "The region of the Redis instance.",
        "examples": [
          "us-central1",
          "asia-northeast1"
        ],
        "pattern": "^[A-Za-z][-a-z0-9A-Z]+$",
        "title": "Region",
        "type": "string"
      }
    },
    "type": "object"
  }
}
```


pkg/broker/service_definition.go
```diff
@@ -129,9 +132,35 @@ func (svc *ServiceDefinition) CatalogEntry() (*Service, error) {
...
+	if enableCatalogSchemas.IsActive() {
+		for i, _ := range sd.Plans {
+			sd.Plans[i].Schemas = svc.createSchemas()
+		}
+	}
+
 	return sd, nil
 }
 
+// createSchemas creates schmas from the provision and bind input variables
```
craigatgoogle: s/schmas/schemas

pkg/broker/service_definition.go
```diff
@@ -129,9 +132,35 @@ func (svc *ServiceDefinition) CatalogEntry() (*Service, error) {
...
+// it populates the instance create schema with the schema generated by
```
craigatgoogle: Awkward wording on this line.

PR #402, opened 2019-01-02
removed individual disable flags

josephlewis42:
Removing the functionality to enable/disable specific services. This was a vestigial artifact from the way services were originally defined in the tile as environment variable JSON blobs.

Builtin services now have a tag and can be enabled/disabled using the tag just like all the other tags.


craigatgoogle:
Out of curiosity, is there anything preventing someone from creating a brokerpak with a service definition using the "builtin" tag? What would be the potential consequences of that?

josephlewis42:
@craigatgoogle that's a really great point. I think I'll change it to a flag on the service definition itself so users can't modify it and it'll keep the tags more pure.

Issue #361, opened 2018-11-29
CVE-2018-15759


josephlewis42:
Our broker uses the library mentioned in: https://pivotal.io/security/cve-2018-15759

The solution is to upgrade the library to version 3.0.2 of the affected library. Because this is a major version bump (currently we're using 2.0.4) and the fix was not back-ported the recommended solution for now is to not run the service broker on a publicly accessible network.

PR #412, opened 2019-01-05
move non-models out of models

josephlewis42:
Removes the last of the non-models out of the models directory, it used to be a kind of general-purpose utility library.

Strings are inlined where they were only used once, otherwise they are moved to the package that owns them.

Finally, removed the differentiation between test and production user agents because the data won't show up in our charts due to our tests using internal projects.


brokerapi/brokers/bigtable/definition.go
```diff
@@ -33,7 +32,7 @@ func ServiceDefinition() *broker.ServiceDefinition {
 	}
 
 	return &broker.ServiceDefinition{
-		Name: models.BigtableName,
+		Name: "google-bigtable",
```
evandbrown: In other definitions where you declare the name in a `const` it seems the only other place it's referenced is in `brokers_test.go`. Is there a reason these service definitions w/ inlined values don't appear in that test?

josephlewis42: That test file is still one of the originals, I think the original author chose to pick a few services that they felt were representative of the cases.

I'm hoping to move it over to just being mocked out today so we can get rid of the need to reference anything by name. That change should also shore up the tests for our main execution path so we can upgrade that vulnerable library that led to #361.

PR #420, opened 2019-01-22
Config Migration

josephlewis42:
This PR adds a config migration package that works like database migrations except for the tile.

Each step of a tile migration can be tested using the builtin JavaScript interpreter otto--which is EMCA5 compatible and is really the lowest common denominator when it comes to JS engines.

While this might seem like overkill right now, there is a large property change coming up affecting default overrides, custom plans, and enable flags.

@craigatgoogle I've been thinking a lot about your comments regarding making this tool equally friendly to use through the CLI as the tile. To this end, the CLI now gets a new migrate-env tool which will do a full migration on environment variables (not config files, unfortunately).

An example of the CLI output is:

```
$ GSB_SERVICE_GOOGLE_BIGQUERY_WHITELIST=foo,bar,bazz ./gcp-service-broker config migrate-env
GSB_SERVICE_GOOGLE_BIGQUERY_WHITELIST:
  old: foo,bar,bazz
  new: ""
```


pkg/generator/pcf-artifacts.go
```diff
@@ -114,6 +111,8 @@ func GenerateTile() string {
 }
 
 func runPcfTemplate(templateString string) string {
+	migrator := migration.FullMigration()
```
mattysweeps: is migrator used?

PR #426, opened 2019-01-25
Validate schema in definition

josephlewis42:
This moves the JSONSchema validation code into the ServiceDefinition where the variables are merged together.

This fixes a weird "bug" where operators setting defaults for required fields would still require the users to enter the values because defaults were applied _after_ the validity of the request was checked.

When fixing this, the tests revealed that the patterns for CloudSQL matching were invalid for the default values so they've been updated to fix that.

This work will eventually feed into the multi-project provisioning because it reduces the responsibilities of the `GCPServiceBroker` struct which has become a bit of a god object. Multi-project provisioning is going to take a significant refactor of that object so it's best to get this out of the way first.


pkg/broker/service_definition.go
```diff
@@ -348,11 +343,28 @@ func (svc *ServiceDefinition) BindVariables(instance models.ServiceInstanceDetai
...
-		Build()
+		MergeDefaults(svc.bindDefaults()).
+		MergeDefaults(svc.BindComputedVariables)
+
+	return buildAndValidate(builder, svc.BindInputVariables)
+}
+
+// buildAndValidate builds the varcontext and if it's valid validates the
+// resulting context against the JSONSchema defined by the BrokerVariables
+// exactly one of VarContext and error will be nil upon return
```
craigatgoogle: Header missing '.' at the end of the sentence.

PR #433, opened 2019-02-06
fixed brokerpak issues

josephlewis42:
- The `pak run-examples` sub-command now returns a non-zero status code on failure.
 - JSONSchema validation no longer fails due to erroneous duplicate required fields.
 - Brokerpaks no longer use incorrect templates due to an invalid pointer.

The intent is to merge this into v4.2.1, then release a hotfix from that branch before merging it back into master

### DIFF
--- a/pkg/broker/variables.go
+++ b/pkg/broker/variables.go
@@ -162,13 +162,13 @@ func ValidateVariables(parameters map[string]interface{}, variables []BrokerVari
 }
 
 func createJsonSchema(schemaVariables []BrokerVariable) map[string]interface{} {
-	var required []string
+	required := utils.NewStringSet()
 	properties := make(map[string]interface{})
 
 	for _, variable := range schemaVariables {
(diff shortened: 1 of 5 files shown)


pkg/broker/variables.go
```diff
@@ -162,13 +162,13 @@ func ValidateVariables(parameters map[string]interface{}, variables []BrokerVari
...
 func createJsonSchema(schemaVariables []BrokerVariable) map[string]interface{} {
-	var required []string
+	required := utils.NewStringSet()
 	properties := make(map[string]interface{})
 
 	for _, variable := range schemaVariables {
 		properties[variable.FieldName] = variable.ToSchema()
 		if variable.Required {
-			required = append(required, variable.FieldName)
+			required.Add(variable.FieldName)
```
mattysweeps: Is there any warnings we should do in case of an override? Or, should we document the behavior that later field definitions override earlier ones?

PR #437, opened 2019-02-11
4-to-5 migration

josephlewis42:
This includes code for the 4 to 5 tile and CLI migrations. As-of v5.0 all configuration for a service will be consolidated to a single environment variable `GSB_SERVICE_CONFIG`.

This variable is a map of service GUID -> service configurations. Each service configuration has a comment, a disabled flag, a list of custom plans, and provision and bind defaults. In total, this consolidates seven forms down to one.

The tile migration takes the following route:

* Whitelist keys are deleted because they're no longer used/supported.
* Custom plans get "collapsed" from something the tile uses internally into a flat JSON map of key->values like what happens when Opsman turns the tile into environment variables.
* Custom plans get formatted, meaning custom properties are put into a format that can be read into a struct without using custom logic in the broker.
* All services with any defined properties get merged into the new config.

What's included:

* A lot of JavaScript.
* A lot of tests.

What's not included:

* The new form field in the tile or documentation for it.

### DIFF
--- a/pkg/config/migration/four-to-five.go
+++ b/pkg/config/migration/four-to-five.go
@@ -43,16 +43,32 @@ func customPlanKeys() (out []string) {
 	return
 }
 
+// MergeToServiceConfig merges the custom plans, enabled, bind defaults, and
+// provision defaults for every 4.x service into a single configuration option.
 func MergeToServiceConfig() Migration {
 	jst := JsTransform{
 		EnvironmentVariables: []string{"GSB_SERVICE_CONFIG"},
 		MigrationJs: `
+
+		function anyExist(envVars) {
+			for(var i in envVars) {
+				if(lookupProp(envVars[i]) !== null) {
+					return true;
+				}
+			}
+
+			return false;
+		}
 
     function mergeService(svcName) {
       var customPlanVar = svcName+"_CUSTOM_PLANS";
       var enabledVar = "GSB_SERVICE_GOOGLE_"+svcName+"_ENABLED";
       var bindDefaultsVar = "GSB_SERVICE_GOOGLE_"+svcName+"_BIND_DEFAULTS";
       var provisionDefaultsVar = "GSB_SERVICE_GOOGLE_"+svcName+"_PROVISION_DEFAULTS";
+
+			if (! anyExist([customPlanVar, enabledVar, bindDefaultsVar, provisionDefaultsVar])) {
+				return null;
+			}
 
       var context = {
         "custom_plans": lookupProp(customPlanVar) || '[]',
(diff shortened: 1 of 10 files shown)


pkg/config/migration/four-to-five.go
```diff
@@ -0,0 +1,246 @@
...
+
+	return
+}
+
+// MergeToServiceConfig merges the custom plans, enabled, bind defaults, and
+// provision defaults for every 4.x service into a single configuration option.
+func MergeToServiceConfig() Migration {
+	jst := JsTransform{
+		EnvironmentVariables: []string{"GSB_SERVICE_CONFIG"},
+		MigrationJs: `
```
craigatgoogle: This makes me so nervous. Sorry if I've asked this before, but is this nested JS code tested somehow?

pkg/config/migration/four-to-five.go
```diff
@@ -0,0 +1,246 @@
...
+	return Migration{
+		Name:       "Merge Service Config",
+		TileScript: jst.ToJs(),
+		GoFunc:     jst.RunGo,
+	}
+}
+
+// DeleteWhitelistKeys removes the whitelist keys used prior to 4.0
+func DeleteWhitelistKeys() Migration {
+	whitelistKeys := []string{
```
craigatgoogle: Any reason not to just in-line this with the function call?

pkg/config/migration/js.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	"bytes"
+	"fmt"
+
+	multierror "github.com/hashicorp/go-multierror"
+	"github.com/robertkrimen/otto"
+)
+
+const jsTransformContext = `
+// env2prop converts an environment variable to property name
```
craigatgoogle: I'm sorry to be complicating things here, but what if you were to move these nested in-line JS blocks into their own files, and loaded them in? That way they could have proper unit tests executing them as a part of your test suite.

pkg/config/migration/four-to-five.go
josephlewis42: The embedded JS should all be tested in the unit-tests. Each test runs the same thing through the go and the JS version of the test. In the case of JsTransform, those use the same logic but different manipulation functions.

In the JS case, the test case is loaded into the `properties` JS field like OpsMan would do, and the test JS is run. Then, the newly modified properties is turned into environment variables via `tile2env` which mimics the way opsman converts the configuration JSON into environment variables.

In the Go case, the test case properties are run through `tile2env` first, then the go version of the function gets run which can either be hand-crafted, or sets up a JS environment which works directly on the environment variable map.

This makes sure that a user running the migration on environment variables gets the same result as if they'd been using the tile directly and exercises both sets of accessor/mutation functions.

I started down the path of having two distinct routes through the code, one written in Go and one in JS, but the code complexity exploded trying to reconcile Tile property names, environment variables, and the different underlying structures. I took a step back and decided that what I really wanted was code that worked through an interface. My "interface" ended up being `lookupProp(envVar) -> (Object|null)`, `deleteProp(envVar)`, and `setProp(envVar, value)`. Those worked cleanly on Go maps, and also on the properties.

The nice part is after the 5.0 release, all of this can be deleted. In the meantime, I'm happy to add anything you think is missing because I know this is less than ideal...

craigatgoogle: Understood, in that case I'm much less nervous about this. Thanks :)

pkg/config/migration/four-to-five.go
josephlewis42: Done

pkg/config/migration/js.go
josephlewis42: That's a good idea, I've now added test-cases specifically for exercising these functions and their go equivalents.

Issue #446, opened 2019-03-06
Support multiple projects


josephlewis42:
The service broker should support provisioning into multiple projects based on rules.

It will do this by letting operators enter a policy to change Terraform environment variables based on labels.

- [x] Expand labels to support more options
- [ ] Add a form for entering a policy
- [ ] Add documentation for the policy
- [ ] Add HIL functions to get an environment variable and coalesce multiple together.
- [ ] Pass these environment variables to TF when it's called.

shinmyung0:
What is the roadmap look like for this? We are running into issues using the gcp service broker for services like Firestore which has a limitation to create only 1 instance per project.

josephlewis42:
Closing this out due to lack of capacity.

PR #447, opened 2019-03-06
added expanded label support

josephlewis42:
Part of #446, expanding supported labels so policies can be written against them for operators to change Terraform environment variables on the fly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
  - A new web-endpoint `/service-config` that hosts service configuration info.
  - Ability to get plan information from HIL execution environment on bind.
  - Ability for plans to selectively override user variables.
+ - Expanded default labels to include org and space name for CF, namespace and clusterid for Kubernetes, and instance-name for all.
+ - A new default label `managed-by:gcp-service-broker` to help identify services which shouldn't be touched.
 
 ### Changed
  - The format of the `/docs` endpoint is now nicely styled with Bootstrap 4.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -21,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
  - Built Brokerpaks now have a name of `{name}-{version}.brokerpak` as defined by the manifest rather than the name of the parent directory.
  - Services inside Brokerpaks now have a file name that includes their CLI friendly name to help differentiate them.
  - The `pak build` command now includes progress logs.
+ - Instance label prefixes are now `cf-` rather than `pcf-`.
```
mattysweeps: Nice catch, It might be worth doing a quick check for any other unnecessary references to pcf, if you haven't already.

docs/billing.md
```diff
@@ -4,9 +4,16 @@ The GCP Service Broker automatically labels supported resources with organizatio
...
- * `pcf-space-guid`
- * `pcf-instance-id`
+* `cf-organization-guid`
+* `cf-organization-name`
+* `cf-space-guid`
+* `cf-space-name`
+* `instance-id`
+* `instance-name`
+* `k8s-namespace`
+* `k8s-clusterid`
```
mattysweeps: Shouldn't `managed-by` also be in this list?

josephlewis42: The `managed-by` is a constant which is why I pushed it into its own section these are all populated from the request.

CHANGELOG.md
josephlewis42: That's a good idea. I see a few more mostly in the default auto-generated names of services. I'll add a ticket to replace these.

PR #449, opened 2019-03-07
added health and liveness checks

### LINKED ISSUE #448: Add health endpoint
The broker needs a health endpoint so helm can do a proper liveness and health check. 

This endpoint should validate access to the GCP project from the provided credentials and database connection.

josephlewis42:
Fixes #448 

I also refactored the serve command as part of this to simplify it and add docs serving from the root.


pkg/server/docs.go
```diff
@@ -65,21 +66,34 @@ var pageTemplate = template.Must(template.New("docs-page").Parse(`
 
 // NewDocsHandler returns a handler func that generates HTML documentation for
 // the given registry.
-func NewDocsHandler(registry *broker.ServiceRegistry) (http.HandlerFunc, error) {
+func NewDocsHandler(router *mux.Router, registry *broker.ServiceRegistry) error {
```
mattysweeps: Since these are no longer initialization functions, can you refactor these to `Add_Handler`?

PR #451, opened 2019-03-07
Remove references to PCF

### LINKED ISSUE #450: Remove references to "pcf-"
We should remove references to "pcf-" and replace them with something that's more inclusive towards CF and K8s.

For example, IAM should go from having a template starting with "pcf-sb-*" to "gcpsb-*"

josephlewis42:
Fixes #450


pkg/providers/builtin/spanner/definition.go
```diff
@@ -70,7 +70,7 @@ func ServiceDefinition() *broker.ServiceDefinition {
 				FieldName: "name",
 				Type:      broker.JsonTypeString,
 				Details:   "A unique identifier for the instance, which cannot be changed after the instance is created.",
-				Default:   "pcf-sb-${counter.next()}-${time.nano()}",
+				Default:   "gsb-${counter.next()}-${time.nano()}",
```
mattysweeps: I just noticed that this uses hyphens where storage/definition.go uses underscores

PR #453, opened 2019-03-19
added retry to delete

josephlewis42:
This should ~fix~ paper over an issue with CloudSQL and Terraform: https://github.com/terraform-providers/terraform-provider-google/issues/1558.

### DIFF
--- a/utils/retrier.go
+++ b/utils/retrier.go
@@ -1,3 +1,16 @@
+// Copyright 2019 the Service Broker Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


utils/retrier.go
```diff
@@ -0,0 +1,34 @@
...
+func Retry(times int, delay time.Duration, callback func() error) error {
+	var err error
+
+	for i := 0; i < times; i++ {
+		err = callback()
+		if err == nil {
+			return nil
+		}
+
+		jitter := time.Duration(rand.Int63n(200)) * time.Millisecond
```
mattysweeps: Can you make Jitter configurable to this function and add it as a constant alongside `DeleteRetryCount` and `DeleteRetryDelay`?

PR #476, opened 2019-06-21
Migrate column request_detail to datatype text

### LINKED ISSUE #468: Provision mysql database fails with "data too long" error
When creating a mysql service with an extensive list of provisioning parameters we get an error as follows:

```
Service broker error: Error saving provision request details to database: Error 1406: Data too long for column 'request_details' at row 1. Services relying on async provisioning will not be able to complete provisioning
```

After a bit of investigation I found a [pull request](https://github.com/GoogleCloudPlatform/gcp-service-broker/pull/267) where @josephlewis42  points out the underlying issue

> Some of our datatypes are suspect, like links only having space for 255 characters when the de-facto standard is ~2000 or RequestDetails which is serialized JSON being a varchar(255)

My opinion is that it doesn't make that much sense to use `varchar(255)` for a column that is supposed to hold json data. 

We should switch the column to text in order to take the marshalled json config.

basiszwo:
closes #468 

Currently the column request_detail in the table ProvisionDetails
is of type string, actually it's varchar(255), which is not enough
space to store JSON based configuration data for creating services.

@josephlewis42 please let me know if I missed something. I am new to service broker development and more than happy to get your feedback and advisory in order to move this thing forward!

### DIFF
--- a/db_service/migrations.go
+++ b/db_service/migrations.go
@@ -67,6 +67,10 @@ func RunMigrations(db *gorm.DB) error {
 		return autoMigrateTables(db, &models.TerraformDeploymentV1{})
 	}
 
+	migrations[5] = func() error { // v4.2.3
+		return autoMigrateTables(db, &models.ProvisionRequestDetailsV2{})
+	}
+
 	var lastMigrationNumber = -1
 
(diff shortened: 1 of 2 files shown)


basiszwo:
I signed it!

PR #485, opened 2019-07-15
Refactor RunExample to run self-contained examples

### LINKED ISSUE #484: Refactor RunExample to run self-contained examples
Refactor RunExample https://github.com/GoogleCloudPlatform/gcp-service-broker/blob/c1bda42985936a4b49e590e0b3227744dbf48c1f/pkg/client/example-runner.go#L60

to change the signature from what it currently is:

```
func RunExample(client *Client, example broker.ServiceExample, service *broker.ServiceDefinition) error {
```

to instead be:

```
func RunExample(client *Client, example CompleteServiceExample) error {

```

and define a new struct `CompleteServiceExample` that includes enough information from the service to run the example (for instance, the current ServiceExamples don't include the service GUID which is necessary if you wanted to execute the example without knowing which service they belonged to)

This will unblock the ability to create a REST endpoint to fetch examples from and feed back to a service-broker.

sophiawho:
This commit resolves #484 

Testing:
- Added example function for `GetAllCompleteServiceExamples` that serves as a sanity check to make sure the function runs error-free, and returns a non-nil array of `CompleteServiceExamples` given the builtin registry.
- Added unit tests to `GetExamplesForAService`, to make sure we could get valid `CompleteServiceExamples` given a `ServiceDefinition`
- Added unit tests for the logic in `FilterMatchingServiceExamples`

### DIFF
--- a/pkg/client/example-runner.go
+++ b/pkg/client/example-runner.go
@@ -33,14 +33,19 @@ import (
 // if serviceName is blank. If exampleName is non-blank then only the example
 // with the given name is run.
 func RunExamplesForService(registry broker.BrokerRegistry, client *Client, serviceName, exampleName string) error {
+
 	rand.Seed(time.Now().UTC().UnixNano())
 
-	allExamples, err := getAllCompleteServiceExamples(registry)
+	if exampleName != "" && serviceName == "" {
+		fmt.Errorf("If an example name is specified, you must provide an accompanying service name.")
+	}
+
+	allExamples, err := GetAllCompleteServiceExamples(registry)
 	if err != nil {
 		return err
 	}
 
-	for _, completeServiceExample := range filterMatchingServiceExamples(allExamples, serviceName, exampleName) {
+	for _, completeServiceExample := range FilterMatchingServiceExamples(allExamples, serviceName, exampleName) {
 		if err := RunExample(client, completeServiceExample); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


pkg/client/example-runner.go
```diff
@@ -55,10 +68,17 @@ func RunExamplesForService(registry broker.BrokerRegistry, client *Client, servi
 
 }
 
+type CompleteServiceExample struct {
+	Example                    broker.ServiceExample   `json: "service_example"`
+	ServiceName                string                  `json: "service_name"`
+	ServiceId                  string                  `json: "service_id"`
+	ServiceBindOutputVariables []broker.BrokerVariable `json: "broker_variables"`
```
josephlewis42: WDYT about renaming this to something like `ExpectedOutput` to better indicate what its purpose is?

pkg/client/example-runner.go
```diff
@@ -55,10 +68,17 @@ func RunExamplesForService(registry broker.BrokerRegistry, client *Client, servi
 
 }
 
+type CompleteServiceExample struct {
+	Example                    broker.ServiceExample   `json: "service_example"`
```
josephlewis42: I think you should embed this reference so all the example's input fields are on the same level.

To get the JSON to be nice, you can add the tag `json:",inline"` which will embed all the child fields at the root level.

pkg/client/example-runner.go
```diff
@@ -45,7 +46,19 @@ func RunExamplesForService(registry broker.BrokerRegistry, client *Client, servi
 				continue
 			}
 
-			if err := RunExample(client, example, service); err != nil {
+			serviceCatalogEntry, err := service.CatalogEntry()
```
josephlewis42: I think it would help to turn this into a function that can construct a complete list of `CompleteServiceExamples` for all the services, then another function that can filter them down to only matching ones before executing them.

Splitting it up like that would make test assertions easier into something like:

* Can we create a correct CompleteServiceExample from some fields?
* Can we create a list of those given a service?
* Can we tell if a single example matches a service and example name?

pkg/client/example-runner.go
sophiawho: Modified struct to:
```
type CompleteServiceExample struct {
	Example        broker.ServiceExample   `json: "service_example, inline"`
	ServiceName    string                  `json: "service_name, inline"`
	ServiceId      string                  `json: "service_id, inline"`
	ExpectedOutput []broker.BrokerVariable `json: "broker_variables, inline"`
}
```

pkg/client/example-runner.go
sophiawho: Good idea! I refactored `RunExamplesForService` and split it up into two functions: `getAllCompleteServiceExamples`, then `runMatchingServiceExamples`. 

Please feel free to check my filtering logic in `runMatchingServiceExamples`.

PR #487, opened 2019-07-23
Add command line flag to enable running examples from file name

### LINKED ISSUE #484: Refactor RunExample to run self-contained examples

sophiawho:
## Description
This commit further resolves #484 by enabling clients to run their own examples against the gcp service broker by providing a `.json` file that contains a list of `CompleteServiceExample` structs. Clients can still specify a specific `serviceName` or `exampleName` along with the file name.

To run the service broker with the file name flag:
```
    ./gcp-service-broker --config config.yml client run-examples --service-examples-file ${FILENAME.json}
```

## Testing
1. An example of how to pretty-print `CompleteServiceExamples` is given in the `ExampleGetAllCompleteServiceExamples2()` func in `example-runner_test.go`.

2. There is a quick sanity check example for running service examples given a file name in the `ExampleRunExamplesFromFile()` func in `example-runner_test.go`. An example file, `complete-service-examples-testfile.json` is given in the same directory.

## Minor Details
1. Moves logic that checks for the error case of a specified `exampleName` and no specified `serviceName` to `client.go` to prevent writing the same logic in two separate functions.
2. Minor indentation nits.


pkg/broker/variables.go
```diff
@@ -161,7 +165,7 @@ func ValidateVariables(parameters map[string]interface{}, variables []BrokerVari
 	return allErrors
 }
 
-func createJsonSchema(schemaVariables []BrokerVariable) map[string]interface{} {
+func CreateJsonSchema(schemaVariables []BrokerVariable) map[string]interface{} {
```
josephlewis42: nit: now that this is public, can we add a docstring to it?

pkg/client/complete-service-examples-testfile.json
```diff
@@ -0,0 +1,25 @@
+[
```
josephlewis42: test fixtures should be put under a `testdata/` directory in the same package as the test--this should be `pkg/client/testdata/complete-service-examples-testfile.json`

It's a very undocumented feature of go: https://medium.com/@povilasve/go-advanced-tips-tricks-a872503ac859

pkg/client/example-runner_test.go
```diff
@@ -17,6 +21,79 @@ func ExampleGetAllCompleteServiceExamples() {
 	// <nil>
 }
 
+func ExampleGetAllCompleteServiceExamples2() {
```
josephlewis42: If you want more than one example against a given type, use an underscore then a camelCase description heading to make it show up in godoc e.g. `ExampleGetAllCompleteServiceExamples_jsonSpec`

pkg/client/example-runner_test.go
```diff
@@ -17,6 +21,79 @@ func ExampleGetAllCompleteServiceExamples() {
...
+	//			"required": [
+	//				"Email"
+	//			],
+	//			"type": "object"
+	//		}
+	//	}
+	//]
+}
+
+func ExampleRunExamplesFromFile() {
```
josephlewis42: Will this work if you aren't running the server?

pkg/client/complete-service-examples-testfile.json
sophiawho: Thanks for letting me know!

sophiawho: Thanks for pointing that out! File moved.

pkg/client/example-runner_test.go
sophiawho: Test removed.

PR #488, opened 2019-07-25
Fix build process

adamwalach:
Changes:
- added .dockerignore file to speed up build process (we don't need .git directory in docker context)
- added ssl certificates to the docker image
- fixed TestBuiltinBrokerRegistry test
- fixed Travis build process


josephlewis42:
Awesome, thanks @adamwalach

Issue #500, opened 2019-07-29
Unable to provision PostgreSQL


piotrmiskiewicz:
The broker is unable to provision PostgreSQL because of too short database column in the internal storage, the log contains an error:
```json
{"timestamp":"1564391052.912573099","source":"gcp-service-broker","message":"gcp-service-broker.provision.unknown-error","log_level":2,"data":{"error":"Error saving provision request details to database: Error 1406: Data too long for column 'request_details' at row 1. Services relying on async provisioning will not be able to complete provisioning","instance-details":{"service_id":"cbad6d78-a73c-432d-b8ff-b219a17a803a","plan_id":"2513d4d9-684b-4c3c-add4-6404969006de","organization_guid":"57230698-b1cf-11e9-89c8-52e15810d407","space_guid":"b92b26f9-b1db-11e9-b50c-42010a84026c","context":{"clusterid":"57230698-b1cf-11e9-89c8-52e15810d407","namespace":"gopher","platform":"kubernetes"},"parameters":{"activation_policy":"ALWAYS","auto_resize":"false","backup_start_time":"06:00","backups_enabled":"true","database_name":"pmiskiew1","disk_size":"10","disk_type":"PD_SSD","instance_name":"pmiskiewsql","maintenance_window_day":"1","maintenance_window_hour":"0","region":"us-central","replication_type":"SYNCHRONOUS","version":"POSTGRES_9_6"},"maintenance_info":{}},"instance-id":"d3bc3f65-b1df-11e9-ac60-0e26a321597b","session":"36"}}
```

The manual change of the MySQL column type to VARCHAR(1024) solved the problem, PostgreSQL was provisioned. It seems the fix is short, just notify gorm about the type in line:
https://github.com/GoogleCloudPlatform/gcp-service-broker/blob/master/db_service/models/historical_db.go#L117

basiszwo:
This was already done in #468. Unfortunately only for new installations. #510 fixes the issue for existing installations.

PR #501, opened 2019-08-02
Set provision_details column type to text

piotrmiskiewicz:
The RequestDetails must be bigger than default 255 varchar for classes with many provisioning parameters like PostgreSQL.

Fixes: https://github.com/GoogleCloudPlatform/gcp-service-broker/issues/500


basiszwo:
This has already been fixed in https://github.com/GoogleCloudPlatform/gcp-service-broker/pull/476 

Should be included in the next release.

josephlewis42:
Closing because I think this has been fixed, we can re-open if that's not the case.

Thanks for the hard work!

PR #502, opened 2019-08-05
 Add new /examples endpoint that returns array of CompleteServiceExamples

### LINKED ISSUE #492: Examples Endpoint
Add a new endpoint to the broker `/examples` that does not require authentication that returns an array of `CompleteServiceExamples` as JSON (`Content-Type: application/json`).

An example of an existing endpoint that returns HTML can be found in `pkg/server/docs.go` it gets attached to the server in `cmd/serve.go`.

Change the client command `run-examples` so it will connect to the server, get the list of examples from the endpoint, and run them. The `--example-name` and `--service-name` flags should continue to work.

An example of how to read JSON from a server in go can be found here: https://blog.alexellis.io/golang-json-api-client/

I recommend taking the following approach:

* Create the endpont and use a web browser to connect to it and validate that it works.
* Create a test for the endpoint similar to `pkg/server/docs_test.go` that validates it returns the correct `Content-Type` and the results can be unmarshaled to the correct JSON type. (maybe split out the logic in `RunExamplesFromFile` that unmarshals examples into a function that unmarshals examples from an `io.Reader` so it'll work with web requests and the file)
* Create a PR with those two pieces
* Refactor the code in the client to fetch examples from the endpoint rather than from the `builtin.BuildtinBrokerRegistry`

sophiawho:
This commit partly resolves #492

### DIFF
--- a/pkg/server/examples.go
+++ b/pkg/server/examples.go
@@ -1,0 +1,46 @@
+// Copyright 2019 the Service Broker Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	"encoding/json"
+	"net/http"
+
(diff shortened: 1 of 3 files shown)


pkg/server/examples.go
```diff
@@ -0,0 +1,46 @@
...
+
+func NewExampleHandler(registry broker.BrokerRegistry) http.HandlerFunc {
+	allExamples, err := client.GetAllCompleteServiceExamples(registry)
+	if err != nil {
+		return func(w http.ResponseWriter, rep *http.Request) {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+		}
+	}
+
+	json, err := json.Marshal(allExamples)
```
josephlewis42: nit: rename the variable `json` so it doesn't shadow the json package, maybe something like exampleJSON

PR #503, opened 2019-08-05
Refactor code in client to fetch examples from the server endpoint

### LINKED ISSUE #492: Examples Endpoint

sophiawho:
This commit resolves #492 

This commit refactors the code in the client to fetch examples from the endpoint rather than from the `builtin.BuildtinBrokerRegistry`. The client command `run-examples` now connects to the server, gets the list of examples from the endpoint, and runs them. The `--example-name`, `--service-name`, and `--filename` flags should continue to work.

Changes include:
- Refactoring some functions from the `client` to `server` pkg (`GetAllCompleteServiceExamples()`, `ExampleGetAllCompleteServiceExamples()` to avoid a circular import dependency
- Re-orders some imports in alphabetical order
- Refactoring code in `brokerpak` pkg to use `GetAllCompleteServiceExamples()` with the pak registry

For testing:
- Ran unit tests in `pkg/client/example-runner_test.go` and `pkg/server/examples_test.go`
- Ran commands to make sure `run-examples` cmd still works with the flags and by itself:
```
go build
./gcp-service-broker serve --config config.yml
./gcp-service-broker --config config.yml client run-examples --service-name google-pubsub
# with the service examples file name flag
./gcp-service-broker --config config.yml client run-examples --filename $HOME/Documents/hsophia-service-examples.json
```

### DIFF
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -21,7 +21,6 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/client"
-	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/providers/builtin"
 	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/server"
 	"github.com/GoogleCloudPlatform/gcp-service-broker/utils"
 )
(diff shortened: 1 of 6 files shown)


cmd/client.go
```diff
@@ -122,13 +125,18 @@ user-defined plans.
 				log.Fatalf("Error creating client: %v", err)
 			}
 
+			allExamples, err := server.GetAllCompleteServiceExamples(builtin.BuiltinBrokerRegistry())
```
josephlewis42: Rather than calling this function here to get allExamples let's create a new function based on the work you did for `TestNewExampleHandler` that will connect to the server, poll the endpoint for the server's list of examples, deserialize them into a list then pass them to `RunExamplesForService`

That way if the list on the server is different than the list the client has then the client will use whatever the server supports.

sophiawho: I created the `GetExamplesFromServer()` function in `pkg/server/examples.go`, based on the tutorial here: https://blog.alexellis.io/golang-json-api-client/.

PR #510, opened 2019-08-07
Add migration to modify column type

basiszwo:
In a previous commit (#bf2000b) the datatype for the column request_details
in the table provision_request_details was changed to text. Unfortunately
this only works for fresh service broker installations as gorm does not
modify the column type for existing columns.

As of this gorm's ModifyColumn has to be used explicitly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
  - Ability for plans to selectively override user variables.
  - Ability to get plan information from HIL execution environment on bind.
 
+### Fixed
+
+- "data too long" error for existing service broker installations when passing service provisioning configuration > 255 characters (#468)
+
 ## [4.2.3] - 2019-06-12
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


basiszwo:
@sophiawho this is in addition to PR #476 to also fix the datatype issue for existing service broker installations.

basiszwo:
The build fails because sqlite does not support changing column types. I refer to https://sqlite.org/lang_altertable.html

maybe changing the test setup to use mysql or postgresql may help.

josephlewis42:
@basiszwo WDYT about adding a condition to check that the database dialect is not sqlite before attempting that statement via `db.Dialect().GetName()`?

sqlite shouldn't care because all columns are just text underneath IIRC

I agree we should have migration tests for MySQL, but I think that's a longer term issue

basiszwo:
@josephlewis42 I totally agree with the fact that switching tests to using mysql is a bigger move. As so, I like your suggestion and I am going to add the dialect check to my code. 

Thank you for guiding me 👍

josephlewis42:
Awesome work, thanks @basiszwo !

PR #525, opened 2019-08-27
Add enum for regions/zones

jasiu001:
In our application we use GCP broker. Our UI helps end-user using simple providing for services. In PR, I propose to use fields of type enum for regions and zone to enhance the general user-experience, manual completion of these fields may cause various types of errors (both on UI and via CLI).

Below is a screen of what the GCP broker looks like together with our application and fields of the type `enum`:

Screen Shot 2019-08-27 at 15 06 06

### DIFF
--- a/pkg/providers/builtin/bigtable/definition.go
+++ b/pkg/providers/builtin/bigtable/definition.go
@@ -16,9 +16,10 @@ package bigtable
 
 import (
 	"code.cloudfoundry.org/lager"
+	"github.com/GoogleCloudPlatform/gcp-service-broker/pkg/broker"
 	accountmanagers "github.com/GoogleCloudPlatform/gcp-service-broker/pkg/providers/builtin/account_managers"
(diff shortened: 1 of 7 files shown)


jasiu001:
@josephlewis42 thank you for a quick answer. Your arguments are really reasonable, I didn't know regions change so often but as I mentioned earlier, leaving the end-user to enter the name of the region is highly prone to errors.

Maybe consider that solution, there will be one list of regions and one list of zones with all regions and zones. User, like is now, must check on the website whether the region for the given service exists but instead copy/paste the region, he can choose one of them from the full list of regions.
Of course, there may be errors when user selects a region from the list that does not exist for the service, but it will be much more user-friendly for him. Does it have any sense for you? 

Another solution could be some endpoint with the matrix for locations for each service, regions could be refreshed e.g. once each day inside the broker, but this solution seems to be labor-intensive and if I'm not mistaken there is no such as endpoint?

jasiu001:
After a discussion in our team, we decided to withdraw the PR. Your explanation shows it was a bad idea.

PR #527, opened 2019-08-29
Cleanup IAM policy when deleting a service account

### LINKED ISSUE #474: Unbind does not delete related IAM policy
### Annotation
**I'm not sure whether it's worth fixing this issue since it does not affect the functionality and the upcoming brokerpak-feature solve this issue by using terraform.**

I more or less just wanted to document it.

### Issue
When binding a service to an app (e.g. CloudSQL), the service broker creates a service account and an IAM policy with the requested role and member.

By unbinding the service, the service broker deletes the service account but leaves the IAM policy behind. This - over time - causes many orphaned IAM policies without any members.

Maybe it does make sense to inform the users to manually delete all orphaned IAM policies when upgrading to the brokerpak?

dominikmueller:
This PR fixes #474.

The IAM policy won't automatically be cleaned up when deleting a service
account. It is necessary to manually remove the service account from all
its role memberships. Otherwise, all created role bindings will remain
existent in the policy, leading to many orphaned bindings.

*Since I'm relatively new to programming in golang:*
I had problems finding an easy solution on how to avoid the code duplication in the methods `revokeRolesFromAccount` and `grantRoleToAccount`.
I would appreciate any hint on how I could/should (or should not?) resolve this! 🙂

### DIFF
--- a/pkg/providers/builtin/account_managers/iam.go
+++ b/pkg/providers/builtin/account_managers/iam.go
@@ -51,3 +51,17 @@ func rolesToMembersMap(bindings []*cloudresourcemanager.Binding) map[string]util
 
 	return bm
 }
+
+// Remove a member from all role bindings to clean up the iam policy before deleting a service account.
+func removeMemberFromBindings(bindings []*cloudresourcemanager.Binding, memberToRemove string) []*cloudresourcemanager.Binding {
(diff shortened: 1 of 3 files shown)


josephlewis42:
Fantastic work @dominikmueller, thanks!

PR #532, opened 2019-09-12
Introduce default expression field to schemas

piotrmiskiewicz:
This PR moves the default HIL expression to new field `Expression` and makes empty default values for such fields. It solves a problem with expressions in default values which does not match required pattern.


piotrmiskiewicz:
The problem was solved by another PR: https://github.com/GoogleCloudPlatform/gcp-service-broker/pull/533

PR #536, opened 2019-09-17
Add HCL 2.0 support for brokerpaks

dominikmueller:
Newer versions of terraform (starting from v0.12) use [HCL 2.0](https://github.com/hashicorp/hcl/wiki/HCL-2.0-Transition) as configuration language.
The new syntax requires new validations and a different parsing of the terraform modules.

HCL 2.0 should be backward compatible with the old syntax and thus should not break the functionality of existing brokerpaks / service definitions.

https://github.com/GoogleCloudPlatform/service-broker-plugins/pull/18 upgrades the existing Service Definitions to terraform v0.12.


josephlewis42:
@dominikmueller have you encountered any issues binding where the broker attempts to read `tfstate` files when moving to Terraform 0.12?

dominikmueller:
I actually encountered multiple issues in my last tests. It seems that with terraform >0.12 there have been made more changes than I expected.

I will take a closer look at those issues and will update this PR accordingly.

dominikmueller:
@josephlewis42 Do you see any issues with switching completely to HCL2.0/terraform 0.12 and removing terraform 0.11 support for the service broker?
Since the brokerpaks are experimental, I would assume that nobody uses them at the moment?

Otherwise, it would be necessary to have different code branches for each terraform version and maybe an abstract provider interface.

Do you have any thoughts about the way to implement such breaking changes in upcoming terraform versions?

PR #545, opened 2019-10-24
Updated CloudSQL with Regional Disks

josephlewis42:


josephlewis42:
@dominikmueller would you mind adding your consent here by leaving a comment that is only `@googlebot I consent.` ?

I expanded your PR by updating the integration tests, and support for CloudSQL for MySQL.

dominikmueller:
@googlebot I consent.

pkg/providers/builtin/cloudsql/mysql-definition.go
```diff
@@ -184,11 +184,11 @@ func MysqlServiceDefinition() *broker.ServiceDefinition {
 			{
 				FieldName: "instance_name",
 				Type:      broker.JsonTypeString,
-				Details:   "Name of the Cloud SQL instance.",
+				Details:   "Name of the CloudSQL instance.",
 				Default:   identifierTemplate,
 				Constraints: validation.NewConstraintBuilder().
 					Pattern("^[a-z][a-z0-9-]+$").
-					MaxLength(84).
+					MaxLength(87).
```
evandbrown: Just confirming this new length is correct.

josephlewis42: nice catch, it looks like it's not

dominikmueller:
Sorry for that wrong max length for the cloudsql instances.
I thought I tried it manually with the google cloud console (but I have no idea why I then got 87 as a max length :/ ).

I think it also needs to be fixed for the postgres instances/definition.

I'm getting the following error message when I try to create an instance in the cloud console:
`The instance name is too long. Must be no more than 83 characters.`

josephlewis42:
@dominikmueller no worries! it looks like that definition is shared to all CloudSQL instances now so I can probably pull the field into common anyway

